Issue #1, opened 2021-01-18
Cannot run the example from the documentation


sheevy:
Hi, I tried to run the example from the documentation but it always fails for me (on both on Windows and Linux and on Julia 1.5.3 and 1.6-beta1).
On freshly installed Julia I run the following (output of most of the commands omitted):
```
(@v1.6) pkg> generate Test2
(@v1.6) pkg> activate Test2
julia> using Pkg
julia> Pkg.add(url="https://github.com/MilesCranmer/SymbolicRegression.jl.git")
julia> using Distributed
julia> addprocs(4)
julia> @everywhere using SymbolicRegression
ERROR: On worker 2:
ArgumentError: Package SymbolicRegression [8254be44-1295-4e6a-a16d-46603ac705cb] is required but does not seem to be installed:
 - Run `Pkg.instantiate()` to install all recorded dependencies.

Stacktrace:
 [1] _require
   @ .\loading.jl:986
 [2] require
   @ .\loading.jl:910
 [3] #1
   @ C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.6\Distributed\src\Distributed.jl:79
 [4] #103
   @ C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.6\Distributed\src\process_messages.jl:274
 [5] run_work_thunk
   @ C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.6\Distributed\src\process_messages.jl:63
 [6] run_work_thunk
   @ C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.6\Distributed\src\process_messages.jl:72
 [7] #96
   @ .\task.jl:406

...and 3 more exceptions.

Stacktrace:
 [1] sync_end(c::Channel{Any})
   @ Base .\task.jl:364
 [2] macro expansion
   @ .\task.jl:383 [inlined]
 [3] _require_callback(mod::Base.PkgId)
   @ Distributed C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.6\Distributed\src\Distributed.jl:76
 [4] #invokelatest#2
   @ .\essentials.jl:707 [inlined]
 [5] invokelatest
   @ .\essentials.jl:706 [inlined]
 [6] require(uuidkey::Base.PkgId)
   @ Base .\loading.jl:916
 [7] require(into::Module, mod::Symbol)
   @ Base .\loading.jl:897
[...]

PR #2, opened 2021-01-18
Clarify the installation instructions

sheevy:
Just in case someone gets puzzled by the problem described there: https://github.com/MilesCranmer/SymbolicRegression.jl/issues/1

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Install in Julia with:
 using Pkg
 Pkg.add(url="https://github.com/MilesCranmer/SymbolicRegression.jl.git")
 ```
-(you need to repeat this for main Julia environment and target working environment (if any))
+(you need to repeat this for both main Julia environment and target working environment (if any))
 
 Run distributed on four processes with:
 ```julia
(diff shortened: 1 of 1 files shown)


MilesCranmer:
Thanks! This is a weird issue I'm seeing too. Let's try again once `SymbolicRegression.jl` is on the main Julia registry, maybe something about the git repo install or the Distributed import messes things up.

Cheers,
Miles

patrick-kidger:
I think the usual pattern for this is something like this:

Install:
```julia
julia> import Pkg
julia> Pkg.activate(...)
julia> Pkg.add(url="https://github.com/MilesCranmer/SymbolicRegression.jl.git")
```
(Or the equivalent from `]`.)

Run:
```julia
import Distributed
Distributed.@everywhere begin
    import Pkg
    Pkg.activate(...)
    using SymbolicRegression
end

...
```

Or:
Run:
```julia
import Distributed

function foo()
    Distributed.@everywhere Main.eval(quote
        import Pkg
        Pkg.activate(...)
        using SymbolicRegression
    end)

    ...
end
```

patrick-kidger:
(Although incidentally I think this Distributed business should be moved inside SymbolicRegression rather than requiring this from the user.)

MilesCranmer:
Thanks @patrick-kidger!! This is very helpful! Will add this to the README.

Cheers,
Miles

PR #3, opened 2021-01-19
Generic types

MilesCranmer:
This converts every `Array{Float32}` into a `AbstractArray{T} ... where {T<:AbstractFloat}`.

Thanks @patrick-kidger for the suggestion and examples!

To-do's remaining:

- [x] Debug if there are any performance changes.
- [x] Are there ways to improve the startup speed? This step seems much slower; perhaps this is inevitable.
- [ ] Add generic types for equation constants.
    - Will do this later; need to check that multiple real types work. Also the user might want different data type to constant type.

Also @ChrisRackauckas, this might be of interest to you since this PR helps open up this package to integration with other SciML packages.

### DIFF
--- a/src/ConstantOptimization.jl
+++ b/src/ConstantOptimization.jl
@@ -3,7 +3,7 @@ using Optim
 # Proxy function for optimization
 function optFunc(x::Vector{ConstantType}, X::AbstractMatrix{T},
                  y::AbstractVector{T}, baseline::T,
-                 tree::Node, options::Options)::T where {T<:AbstractFloat}
+                 tree::Node, options::Options)::T where {T<:Real}
     setConstants(tree, x)
     return scoreFunc(X, y, baseline, tree, options)
 end
@@ -11,7 +11,7 @@ end
 # Use Nelder-Mead to optimize the constants in an equation
 function optimizeConstants(X::AbstractMatrix{T}, y::AbstractVector{T},
                            baseline::T, member::PopMember,
-                           options::Options)::PopMember where {T<:AbstractFloat}
+                           options::Options)::PopMember where {T<:Real}
 
     nconst = countConstants(member.tree)
     if nconst == 0
--- a/src/ConstantOptimization.jl
+++ b/src/ConstantOptimization.jl
@@ -1,19 +1,24 @@
 using Optim
 
 # Proxy function for optimization
-function optFunc(x::Array{Float32, 1}, X::Array{Float32, 2}, y::Array{Float32, 1}, baseline::Float32, tree::Node, options::Options)::Float32
+function optFunc(x::AbstractArray{T, 1}, X::AbstractArray{T, 2},
+                 y::AbstractArray{T, 1}, baseline::T,
+                 tree::Node, options::Options)::Float32 where {T<:AbstractFloat}
     setConstants(tree, x)
     return scoreFunc(X, y, baseline, tree, options)
 end
 
 # Use Nelder-Mead to optimize the constants in an equation
-function optimizeConstants(X::Array{Float32, 2}, y::Array{Float32, 1}, baseline::Float32, member::PopMember, options::Options)::PopMember
+function optimizeConstants(X::AbstractArray{T, 2}, y::AbstractArray{T, 1},
+                           baseline::T, member::PopMember{T},
+                           options::Options)::PopMember{T} where {T<:AbstractFloat}
+
     nconst = countConstants(member.tree)
     if nconst == 0
         return member
     end
     x0 = getConstants(member.tree)
-    f(x::Array{Float32,1})::Float32 = optFunc(x, X, y, baseline, member.tree, options)
+    f(x::AbstractArray{T, 1})::T = optFunc(x, X, y, baseline, member.tree, options)
     if size(x0)[1] == 1
         algorithm = Newton
     else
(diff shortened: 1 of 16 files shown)


src/Equation.jl
```diff
@@ -147,12 +147,12 @@ end
 
 
 # Get all the constants from a tree
-function getConstants(tree::Node)::Array{Float32, 1}
+function getConstants(tree::Node)::AbstractArray{ConstantType, 1}
```
ChrisRackauckas, line 150: It's common to use `T` for the type in this kind of situation.

src/LossFunctions.jl
```diff
@@ -1,45 +1,47 @@
 using Random: randperm
 
 # Sum of square error between two arrays
-function SSE(x::Array{Float32}, y::Array{Float32})::Float32
+function SSE(x::AbstractArray{T}, y::AbstractArray{T})::T where {T<:AbstractFloat}
```
ChrisRackauckas, line 4: `AbstractFloat` seems a little tighter than necessary for a lot of these. If you want to put a bound on it, I think `Real` might be the right level. For example, rational numbers should be fine.

src/MutationFunctions.jl
```diff
@@ -30,8 +30,8 @@ function mutateConstant(
         node = randomNode(tree)
     end
 
-    bottom = 0.1f0
-    maxChange = options.perturbationFactor * T + 1.0f0 + bottom
+    bottom = convert(T, 0.1)
```
ChrisRackauckas, line 33: `0.1` does not have an example Float64 representation, so it might be worthwhile to have this as an exact constant `1//10` which converts to the best representation in the given type. Very minor detail though. IIRC the compiler should specialize on that during method specialization and essentially convert it before runtime.

ChrisRackauckas:
Looks great! I left a few minor comments about generic coding style just because I couldn't help it. For a symbolic package like this it mostly doesn't matter though since I don't know how many people will give you >64bit precision data to do symbolic regression on (though I've found if you've made things generic, the people will come. Neural boundary value problems with quaternions is my favorite example haha).

I think we'll need to find a way to connect this to SymbolicUtils.jl. For reference, ModelingToolkit is splitting out its CAS portions into a Symbolics.jl which is a full "Sympy-like" system. ModelingToolkit itself will then be kept as a system modeling package (in a sense similar to the Modelica language if you're familiar with that). DataDrivenDiffEq.jl's SINDy I think has already refactored to separate out a `symbolic_regression` function (@AlCap23). It's built on the "ModelingToolkit representation" which will shortly become the Symbolics.jl representation when that split occurs. 

Pinging @yingboma @shashi so they are aware of this.

ChrisRackauckas:
> Are there ways to improve the startup speed? This step seems much slower; perhaps this is inevitable.

The only difference might be that it's not precompiling the right dispatches, so you might want to add a SnoopCompile to get that back. I think there's some new tooling that automates this process, though I haven't dug into it quite myself yet.

AlCap23:
Nice! Sorry to dive in here, but I am happy right now :). Feel free to ignore this. Full disclosure, I've peeked into your Python Code some time ago.

I've been working on an implementation of FFX, but some other things slowed that down. The building blocks are somewhat there except for mutation, constraints in terms of the tree and generic optimization routines. If you're interested, I would love to have a chat on slack or per email to discuss your work here and possible use inside DataDriven.

src/Equation.jl
patrick-kidger: I think `ConstantType` is meant here in the sense of a strong type / named type. I don't think `T` would be a very a meaningful name?

ChrisRackauckas: Yeah I noticed later he was using T. I just saw this early and was like, is he using `ConstantType` all the way though? No he didn't, so this can be ignored.

MilesCranmer:
Awesome, thanks very much for the review @ChrisRackauckas!! It is very helpful to know these tips.

I will implement these suggestions.

Sounds good. It would be fantastic to incorporate integration to SymbolicUtils.jl. SymbolicRegression.jl has its own internal data structure for equations---a binary tree with an `enum` type for selecting which operator to use---this makes evaluation of runtime-generated equations very fast with very few memory allocations. I would really like to have converters in both ways for equations - that would enable SymbolicRegression.jl to use things like simplification routines internally, and also enable SymbolicUtils.jl to use the symbolic regression routines here. (I have some basic hand-written simplification strategies but they are extremely limited.)

@AlCap23, that sounds amazing. Let's definitely have a chat!! I would be very interested in working with you on an FFX implementation! My email is miles.cranmer@gmail.com - want to ping me with your email?

Cheers,
Miles

ChrisRackauckas:
I think that it makes sense to use a problem-specific representation here internally for this because of the specific requirements for speed on a growing equation. But indeed, what I was looking for was a way to convert the inputs/outputs across. If that exists, I think everything else falls into place pretty easily.

src/ConstantOptimization.jl
```diff
@@ -1,19 +1,24 @@
 using Optim
 
 # Proxy function for optimization
-function optFunc(x::Array{Float32, 1}, X::Array{Float32, 2}, y::Array{Float32, 1}, baseline::Float32, tree::Node, options::Options)::Float32
+function optFunc(x::Vector{ConstantType}, X::AbstractMatrix{T},
```
patrick-kidger, line 4: I think the `Abstract` has been left off here...

src/ConstantOptimization.jl
```diff
@@ -1,19 +1,24 @@
...
+                           baseline::T, member::PopMember,
+                           options::Options)::PopMember where {T<:Real}
+
     nconst = countConstants(member.tree)
     if nconst == 0
         return member
     end
     x0 = getConstants(member.tree)
-    f(x::Array{Float32,1})::Float32 = optFunc(x, X, y, baseline, member.tree, options)
+    f(x::Vector{ConstantType})::T = optFunc(x, X, y, baseline, member.tree, options)
```
patrick-kidger, line 21: ...and here.

src/ConstantOptimization.jl
MilesCranmer: I think these are okay because these are fixed internal types. I use this vector to get/set the constants in an equation when doing optimization. Thoughts?

MilesCranmer:
Sounds good! I will work on a converter. I think output from SymbolicRegression.jl to SymbolicUtils.jl will be very easy - I just need to run an eval() on `stringTree` after declaring the variables with `@syms`.

Going the other way will be a bit more work but I think is doable.

src/ConstantOptimization.jl
patrick-kidger: If it's all internal then I agree it's probably unimportant.

MilesCranmer:
I don't detect performance changes, so merging this now.

Thanks for taking a look through the changes!

Cheers,
Miles

PR #4, opened 2021-01-19
Symbolic utils converter

MilesCranmer:
@ChrisRackauckas @patrick-kidger @YingboMa @shashi @AlCap23

Here is a first draft of a PR to work on a converter from SymbolicRegression.jl to SymbolicUtils.jl. See `src/InterfaceSymbolicUtils.jl`.

**Edit:** the technique traverses my equation data structure, and constructs a symbolic expression as it goes using a pre-defined list of variables in `SymbolicUtils.jl` syntax.

By the way, is there a dynamic way to declare symbols with function-dependent type? I wasn't sure if there is a dynamic way to set this up: the current way is quite rough.

For example:

```julia
using Distributed
using Random
using SymbolicRegression

X = randn(Float32, 100, 5)
y = 2 * cos.(X[:, 4]) + X[:, 1] .^ 2 .- 2

options = SymbolicRegression.Options(
    binary_operators=(+, *),
    unary_operators=(cos, exp),
    npopulations=20
)

Random.seed!(0)

# Random population of equations:
pop = Population(X, y, 1f0, 100, options, 5)

println(to_symbolic(X, pop.members[1].tree, options))
```

Output:
```julia
exp((x4 * x4) * 0.1215901)
```


MilesCranmer:
I cleaned things up. Now, instead of I explicitly construct the SymbolicUtils.jl expression by traversing my package's internal tree structure. This should be much more robust.

PR #6, opened 2021-01-29
Dimension swap

MilesCranmer:
Following Flux.jl's format, this PR switches the batch axis to be the last axis. Now the input `X` array is [features, rows] rather than previous [rows, features].

I have `@deprecated` `RunSR` to do this transpose automatically, so existing code should still work. I also created a new function `EquationSearch` which requires the proper format at input. This will be the main SR function going forward.

I do not plan on changing the array format of https://github.com/MilesCranmer/PySR - that one will remain [rows, features] following the standard format in many Python ML frameworks. I will just have https://github.com/MilesCranmer/PySR/pull/26 transpose the array when calling out to the Julia backend.

@patrick-kidger how does this look?

cc @DhananjayAshok

### DIFF
--- a/src/Deprecates.jl
+++ b/src/Deprecates.jl
@@ -1,0 +1,4 @@
+using Base: @deprecate
+
+# Now the batch dimension is the last axis!
+@deprecate RunSR(X, y; kw...) EquationSearch(convert(Array, X'), y; kw...)
--- a/src/Deprecates.jl
+++ b/src/Deprecates.jl
@@ -1,4 +1,4 @@
 using Base: @deprecate
 
 # Now the batch dimension is the last axis!
-@deprecate RunSR(X, y; kw...) EquationSearch(convert(Array, X'), y; kw...)
+@deprecate RunSR(X, y; kw...) EquationSearch(copy(transpose(X)), y; kw...)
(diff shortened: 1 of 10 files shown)


src/Deprecates.jl
```diff
@@ -0,0 +1,4 @@
+using Base: @deprecate
+
+# Now the batch dimension is the last axis!
+@deprecate RunSR(X, y; kw...) EquationSearch(convert(Array, X'), y; kw...)
```
patrick-kidger, line 4: `X'` is the adjoint, not the transpose, of a matrix. Subtly different things.
Also I don't think the `convert` should be necessary provided `EquationSearch` takes the argument type to be `AbstractArray`.

patrick-kidger (commented):
Nitpick aside, LGTM!

src/Deprecates.jl
MilesCranmer: Ah, good catch! I guess it doesn't matter since we enforce `<:Real` types but still probably good to change since complex numbers would be nice to eventually include.

I convert to Array since I wasn't sure if `X'` or `transpose(X)` is just a new view of the array, or if it actually reshapes it? For example:

```julia
X = randn(Float32, 100, 5);
y = transpose(X)
typeof(y)
```
gives: 
```
LinearAlgebra.Transpose{Float32,Array{Float32,2}}
```
any idea?

Thanks!
Miles

MilesCranmer: Oh, it seems like `copy(y)` is enough to force it to change the shape

MilesCranmer: Okay, pushed!

PR #7, opened 2021-01-30
Automatically configure workers internally

MilesCranmer:
cc @patrick-kidger

EquationSearch now takes two new optional kwargs:
- `numprocs` which allows one to set the number of workers to set up
- `procs`, which allows one to explicitly pass the indices of worker nodes.

The `procs` argument allows one to run the equation search on a set part of a cluster, so that they are not automatically distributed over all nodes. The `numprocs` basically makes it easier to do multiprocessing - one doesn't need to use `@everywhere` when setting up the code.

Example:
```julia
include("src/SymbolicRegression.jl")
using .SymbolicRegression
using SymbolicUtils

X = randn(Float32, 5, 100)
y = 2 * cos.(X[4, :]) + X[1, :] .^ 2 .- 2

inv(x) = 1/x

options = SymbolicRegression.Options(
    binary_operators=(+, *),
    unary_operators=(cos, exp, inv),
    npopulations=2
)
niterations = 2

hallOfFame = EquationSearch(X, y, niterations=niterations, options=options, numprocs=4)

dominating = calculateParetoFrontier(X, y, hallOfFame, options)
eqn = node_to_symbolic(dominating[end].tree, options, evaluate_functions=true)

print(simplify(eqn*5 + 3)) # simplify with SymbolicUtils.jl
```

This is a draft PR right now because I can't quite get it to work... any idea what this bug is from @patrick-kidger?
```
ERROR: LoadError: On worker 2:
UndefVarError: SymbolicRegression not defined
deserialize_module at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Serialization/src/Serialization.jl:956
...
...
...
 [1] sync_end(::Channel{Any}) at ./task.jl:314
 [2] macro expansion at ./task.jl:333 [inlined]
 [3] EquationSearch(::Array{Float32,2}, ::Array{Float32,1}; niterations::Int64, weights::Nothing, varMap::Nothing, options::Options{Tuple{typeof(+),typeof(*)},Tuple{typeof(cos),typeof(exp),typeof(inv)}}, numprocs::Nothing, procs::Nothing) at /Users/mcranmer/SymbolicRegression.jl/src/SymbolicRegression.jl:137
 [4] top-level scope at /Users/mcranmer/SymbolicRegression.jl/example.jl:17
 [5] include(::Function, ::Module, ::String) at ./Base.jl:380
 [6] include(::Module, ::String) at ./Base.jl:368
 [7] exec_options(::Base.JLOptions) at ./client.jl:296
 [8] _start() at ./client.jl:506
in expression starting at /Users/mcranmer/SymbolicRegression.jl/example.jl:17
```


MilesCranmer:
@ChrisRackauckas do you know of a good way to initialize all module methods on workers which get created at runtime *within* said module?

patrick-kidger:
So I'm having a look at this now.

Independently of this, I would note that you also need to activate the current environment on each worker, as they don't automatically inherit the current environment. I'm not sure how to get the currently activated environment programatically, though once you have actually activating it is easy, `Pkg.activate(...)`.

patrick-kidger:
Side note, I don't think the nested `@async @spawnat ...` is really necessary? It looks like `@spawnat` returns a future, and indeed the following prints only 2 seconds:

```julia
using Distributed

procs = addprocs(10)

@time begin
   futures = []
   for proc in procs
       push!(futures, @spawnat proc sleep(2))
   end
   for fut in futures
       fetch(fut)
   end
end
```

Anyway, I've managed to produce the following MWE for the issue you're describing:
```julia
module A
    using Distributed

    procs = addprocs(10)

    begin
       futures = []
       for proc in procs
           push!(futures, @spawnat proc sleep(2))
       end
       for fut in futures
           fetch(fut)
       end
   end
end
```
```
ERROR: On worker 6:
UndefVarError: A not defined
```

So it's definitely just the encapsulation inside the module that's producing funny behaviour, not whatever it is that you're running.

MilesCranmer:
Thanks, good catch on the `@async` redundancy.

Yeah, this seems like a tricky problem. On the discourse I can't find any working example of adding a user-defined number of procs inside a module. Maybe it is just standard to always set up the procs on the user's end, and pass those procs to the function call....

MilesCranmer:
Also, I should probably switch to using this function instead of just calling the same `include` on the worker nodes, since the path might be different for workers: https://github.com/ChrisRackauckas/ParallelDataTransfer.jl/blob/07333b5693e812ea283b969ae8f9e887098a3e69/src/ParallelDataTransfer.jl#L80

MilesCranmer:
Okay, it actually looks looks doable using this package Reexport, e.g., they addprocs inside a module here: https://github.com/USCqserver/HPCManager.jl/blob/84d77316cee9794dbcbc03e22fda29af3816e7ed/src/HPCManager.jl

https://github.com/simonster/Reexport.jl

MilesCranmer:
Hm, doesn't seem to work (hangs).

Maybe this one: https://github.com/CliMA/GriddingMachine.jl/blob/e7584c1adf90b20d6b80cb131c74ecb78933f93f/src/Gridding/workers.jl#L34

patrick-kidger:
> since the path might be different for workers

In a multi-computer distributed setup? Makes sense, I think.

> doable using this package Reexport

I think that's unrelated. They're just making `HPCManager` also provide everything that `Distributed` does, plus all of their stuff, as a convenience.

MilesCranmer:
I tried the following code inside EquationSearch - it seems to run (though I guess we should only run this if `need_remove_procs` is set `true` (meaning SymbolicRegression.jl created the processes):
```julia
    @everywhere procs begin
        Base.MainInclude.eval(include(joinpath(@__DIR__, "SymbolicRegression.jl")))
        Base.MainInclude.eval(using .SymbolicRegression)
    end
```
and it doesn't complain about the module anymore. These actually execute okay.

However, after this, after the `@spawnat`'s launch, it seems to hang. So my guess is there's an error happening inside the workers somewhere...

MilesCranmer:
@patrick-kidger I got the minimal working example working! See this code:

`mwe.jl`
```julia
module A
	export test
	using Distributed

	function test(procs)
		futures = []
		for proc in procs
			push!(futures, @spawnat proc println("Hello from $proc"))
		end
		for fut in futures
			fetch(fut)
		end
	end

	function test()
		procs = addprocs(10)
		@everywhere procs begin
			Base.MainInclude.eval(include(@__FILE__))
			Base.MainInclude.eval(using .A)
		end
		test(procs)
	end
end
```

`run.jl`
```julia
include("mwe.jl")
using .A
test()
```

Output, despite the user script not importing `Distributed`:
```
      From worker 9:    Hello from 9
      From worker 6:    Hello from 6
      From worker 10:   Hello from 10
      From worker 7:    Hello from 7
      From worker 3:    Hello from 3
      From worker 2:    Hello from 2
      From worker 4:    Hello from 4
      From worker 5:    Hello from 5
      From worker 11:   Hello from 11
      From worker 8:    Hello from 8
```

patrick-kidger:
Amazing!

Meanwhile, I'm happy to say that I've figured out how to make projects/environments work.

Putting them together, I've tested that the following works when installed as a package called `SymbolicRegression`:
```julia
# SymbolicRegression.jl

module SymbolicRegression

import Distributed
import Pkg

function test(procs)
    # Set up each worker process
    project_path = splitdir(Pkg.project().path)[1]
    Distributed.@everywhere procs begin
        Main.eval(quote
            import Pkg
            Pkg.activate($$project_path)
            import SymbolicRegression
        end)
    end

    Distributed.pmap(_->println("Hello from " * string(Distributed.myid())), Distributed.WorkerPool(procs), 1:6)
end

function test(num_procs::Integer)
    project_path = splitdir(Pkg.project().path)[1]
    procs = Distributed.addprocs(num_procs)
    try
        test(procs)
    finally
        Distributed.rmprocs(procs)
    end
end

end
```

This seems to be extremely fragile. If  `@everywhere` is replaced with `@spawnat` it doesn't work. Curiously if `import SymbolicRegression` is replaced with your use of `MainInclude`, it also doesn't work. I don't know if that's related to being installed as a package or not.

MilesCranmer:
Actually I think the `@__FILE__` and `using .SymbolicRegression` should already do this, no? Because `@__FILE__` should locate the version of the code it has imported - even if it's inside a project environment? (it should be the same file used by the head node)

I pushed a commit to implement this in the module - it seems to work! (at least with a local install).

However, one big issue is I can't seem to pass operators that are defined by the user to the workers (i.e., not defined by the package). Will keep looking at this...

MilesCranmer:
Okay, it seems like it is REALLY hard to pass functions to workers, where the function symbol is passed by the user to a function, and the function is only defined on the head node.

This is the closest I've gotten: https://discourse.julialang.org/t/easy-way-to-send-custom-function-to-distributed-workers/22118/2, but I just can't seem to replace `my_func` with the user-passed operator name:

```julia
        for op in (options.binops..., options.unaops...)
            @everywhere procs @eval function $op end
            #^ this part fails.It works if I write the function explicitly.
            for m in methods(op)
                @everywhere procs @eval $m
            end
        end
```
I've `@macroexpand`ed so many different macros in Distributed, but cannot for the life of me get this to work...

I've even started to try to use `Base.uncompressed_ast(first(methods(op)))` to get the source for a function, and send that to workers to evaluate. Maybe that's a promising route but it feels a bit overkill. However I can't even get that to run...

patrick-kidger:
In response to your first comment: I think there's a slight difference in that my approach involves loading `SymbolicRegression` as a package, whilst yours makes it a local module. I think that should mean mine takes advantage of precompilation whilst yours doesn't, but beyond that I don't know if it's important.

For the second point, I don't have any issue. Modifying by example above like so:

```julia
module SymbolicRegression

import Distributed
import Pkg

function test(procs, fn)
    # Set up each worker process
    project_path = splitdir(Pkg.project().path)[1]
    Distributed.@everywhere procs begin
        Main.eval(quote
            import Pkg
            Pkg.activate($$project_path)
			import SymbolicRegression
        end)
    end

    Distributed.pmap(_->println("Hello with value " * string(fn(Distributed.myid()))), Distributed.WorkerPool(procs), 1:6)
end

function test(num_procs::Integer, fn)
    project_path = splitdir(Pkg.project().path)[1]
    procs = Distributed.addprocs(num_procs)
    try
        test(procs, fn)
    finally
        Distributed.rmprocs(procs)
    end
end

end

# On the REPL
import SymbolicRegression
SymbolicRegression.tests(2, x->x)
```

MilesCranmer:
> takes advantage of precompilation

Ah, good call, right! Will change it.

Regarding the example you posted, two comments:

1. I'm not sure if this is true, but passing `x->x` might be different than writing:
```julia
id(x) = x
SymbolicRegression.test(2, id)
```
But again, not sure if passing `x->x` will include more information for the workers or not...
2. Even so, I can't actually run your example. I get this error when running from the REPL:
```julia
ERROR: On worker 2:
UndefVarError: #1#2 not defined
```
Which is the same error I get when trying to pass functions to workers normally. Not sure how you are specifically running it, but I think the workers need to explicitly be passed the function somehow.


---

I think the example here: https://discourse.julialang.org/t/easy-way-to-send-custom-function-to-distributed-workers/22118/2 should be enough to do it. I just can't figure out the right combination of `@eval`, `$` and `:()` to get it working...

i.e., this is fine:
```julia
@everywhere workers() @eval function my_func end
```

but this gives an error:
```julia
tmp = my_func
@everywhere workers() @eval function $tmp end
```
error:
```
ERROR: On worker 2:
UndefVarError: #my_func not defined
```

This is driving me crazy but I can't seem to figure out the difference, even after exploring the `@macroexpand`s.

This is the closest I could get by macro expanding, and making sure that the correct function name is inside the expression *before* being passed to the worker:
```julia
tmp2 = :(function $tmp end)
let tmp3 = Expr(:copyast, :($(QuoteNode(:(@eval($tmp)))))), tmp4 = workers()
    Distributed.remotecall_eval(Distributed.Main, tmp4, tmp3)
end
```
but this gives me this strange error:

```
ERROR: LoadError: type GeneratedFunctionStub has no field inferred
```

MilesCranmer:
Okay I got it working. The code is ugly but now it seems to actually run with a custom operator!

The two fixes: need to pass the name of the operator to the `@eval`, and also should avoid re-defining existing operators - hence the `catch` being used. I think my re-defining of existing operators was causing the error above.

```julia
        for degree=1:2
            ops = degree == 1 ? options.binops : options.unaops
            for op in ops
                try
                    if degree == 1
                        fetch(@spawnat next_worker() op(convert(T, 0)))
                    else
                        fetch(@spawnat next_worker() op(convert(T, 0), convert(T, 0)))
                    end
                catch
                    name = nameof(op)
                    src_ms = methods(op).ms
                    @everywhere procs @eval function $name end
                    for m in src_ms
                        @everywhere procs @eval $m
                    end
                end
            end
        end
```

MilesCranmer:
Followup question: do you know how to detect whether the user is importing via 
```using .SymbolicRegression``` or
```using SymbolicRegression```? Since we likely don't want to import the installed version in worker nodes if the head node is using an included local version (though I guess this just applies to developers of this package...).

**Comment by patrick-kidger** [comment]
1. There's slight differences, I think: the first one introduces a name into the current scope, after all. I also don't know if that matters.

2. The example in my last comment works when used as a package. Indeed I'm not surprised that it won't work from the REPL (although curiously I get a different error to you when I try). Somehow we need to specify the code to send to the worker processes, and everything we've been doing so far has been predicated on that code existing in the file system somewhere -- whether accessed via `include` (as per some time ago now) or `import` (as per above).

The above also applies to your last comment. In principle I think one could try looking in `Base.loaded_modules`, but I think that'll probably be too much magic, and will end up biting us in some edge case.

I'm quite skeptical about both the try/catch and the eval'ing of functions. It seems complicated and error-prone. For example it will trigger if the user-provided op raise an error. Additionally in the catch branch, why do you define a zero-argument version of the function, and then define all its actual methods afterwards?

**Comment by patrick-kidger** [comment]
By the by, it'd be worth thinking about whatever we do also making sense in a single-process-single-thread or single-process-multi-threaded context. I assume you explicitly need distributed operation (over threaded operation) or else you wouldn't be doing this, but I think most applications I'd end up using this for would prefer to keep this on one machine with a threaded approach.

**Comment by MilesCranmer** [comment]
1. I think it's actually better if anonymous functions are disallowed here, purely because the user wouldn't be able to recognize them in the printed-out equations anyway! The operators just get displayed like `#1`. Maybe I should add an error for this.
2. Great points. I think `Base.loaded_modules` is a good idea - it seems like that's what Julia checks for existing packages anyway to see if there is a re-definition. 
3. (re: try/catch) - I just added code to check that the code gives `UndefVarError`, which is what we are looking for. It's definitely complex to get this working, but it just doesn't seem like adding procs within a module is very well supported, so I guess we are trail-blazing a bit... 
> Additionally in the catch branch, why do you define a zero-argument version of the function, and then define all its actual methods afterwards?
- So the `func myfunc end` version is basically like initializing the symbol without any methods defined. In particular, it's **not** a zero-argument method; that would be `func myfunc() end`, which is (confusingly) both the method + function definition - the method returning `nothing` to a call with zero arguments. I guess if one wants to manually add methods like this, one needs to first initialize the symbols like this. (from here https://discourse.julialang.org/t/easy-way-to-send-custom-function-to-distributed-workers/22118/2).
4.
- Re: single-process multi-thread; I'll note there's no performance difference here - I benchmarked this when I first started out. Perhaps if I had a large amount of shared data, there would be (I guess this is why multi-threading is more common in other HPC stuff), but the datasets expected for symbolic regression are small enough that one can just copy them. And equations aren't copied outside of each "migration period," so there shouldn't be a hit there either.
- The main reason it uses Distributed is because it's effortless to move between a single machine and multi machine via ClusterManagers: https://github.com/JuliaParallel/ClusterManagers.jl - with this you can add procs across a cluster with the same command as used for single machine! But with `Threading`, one would need to mix both threads and processes which might complicate things.

**Comment by patrick-kidger** [comment]
1. Good point.

2.,3. Obviously it's your package and far be it from me to tell you what to do -- but the stuff you've posted looks crazy. The last code snippet I posted works just fine! 

I think working with `Base.loaded_modules` or try-catching anything is going to silently go wrong in unpredictable ways.

That said I agree that it seems we're trailblazing here.

(unnumbered Re: `function name end`) Huh, I learnt something new today.

4. What you say makes sense. Incidentally you may like to know that `Distributed.pmap` offers a single interface to both processes and threads, if that's ever useful in some other context.

**Comment by MilesCranmer** [comment]
2,3 - completely agree the code looks crazy! I'm just not sure how to pass functions to the workers otherwise though (and allow local installs). I did add your snippet for normal package activation - good call! It all does seem to work, at the expense of internal complexity.

Here's two functions to set up the module on workers, called from EquationSearch:

```julia
function activate_env_on_workers(procs, project_path)
    println("Activating environment on workers.")
    @everywhere procs begin
        Base.MainInclude.eval(quote
            using Pkg
            Pkg.activate($$project_path)
        end)
    end
end

function import_module_on_workers(procs, filename::String)
    included_local = !("SymbolicRegression" in [k.name for (k, v) in Base.loaded_modules])
    if included_local
        println("Importing local module ($filename) on workers.")
        @everywhere procs begin
            # Parse functions on every worker node
            Base.MainInclude.eval(quote
                include($$filename)
                using .SymbolicRegression
            end)
        end
    else
        println("Importing installed module on workers.")
        @everywhere procs begin
            Base.MainInclude.eval(using SymbolicRegression)
        end
    end
end
```

4. This is good to know! I actually wonder how much would *just work* if we swap the `Distributed` module with `Threads`...  (and `@spawnat w` -> `@spawn`, etc.) 


5. By the way, here are the full functions to move functions to workers. It's ugly but it seems to work. There's a lot of initialization tests now, so any issues on the workers should be more obvious, instead of just having things hang.

```julia
# Re-define user created functions on workers
function move_functions_to_workers(T, procs, options::Options)
    for degree=1:2
        ops = degree == 1 ? options.unaops : options.binops
        for op in ops
            try
                test_function_on_workers(T, degree, op, procs)
            catch e
                undefined_on_workers = isa(e.captured.ex, UndefVarError)
                if undefined_on_workers
                    name = nameof(op)
                    println("Copying definition of $op to workers.")
                    src_ms = methods(op).ms
                    # Thanks https://discourse.julialang.org/t/easy-way-to-send-custom-function-to-distributed-workers/22118/2
                    @everywhere procs @eval function $name end
                    for m in src_ms
                        @everywhere procs @eval $m
                    end
                else
                    throw(e)
                end
            end
            # Make sure it actually worked!
            test_function_on_workers(T, degree, op, procs)
        end
    end
end

function test_function_on_workers(T, degree, op, procs)
    futures = []
    for proc in procs
        if degree == 1
            push!(futures,
                  @spawnat proc op(convert(T, 0)))
        else
            push!(futures,
                  @spawnat proc op(convert(T, 0), convert(T, 0)))
        end
    end
    for future in futures
        fetch(future)
    end
end

```

MilesCranmer:
Okay, this is weird: the `try...finally..end` seems to  hurt the performance by like 5x!! I was worrying if creating procs within the module might have affected it, but commenting out the try-finally statement brought it back up to speed. 

After Googling, this seems to be generally the case for listening for exceptions in Julia - I had no idea! Will check if I have any other try statements encapsulating important chunks of code.

MilesCranmer:
Okay, this seems to be working pretty well, both when ran from a local file, and from an environment compiled version!

PR #13, opened 2021-02-10
Use FromFile.jl's `@from ... import` syntax instead of include()

MilesCranmer:
Switches SymbolicRegression.jl to use the explicit dependency system of https://github.com/Roger-luo/FromFile.jl rather than `include()`.

For loading modules on the workers, I did need to do an explicit "include" of the `Configure.jl` file, for whatever reason. But everything else seems to work fine and I can run the pipeline normally.

@patrick-kidger what do you think?

The file system has been restructured as follows. Indentation
shows dependencies. I had to move some of the equation utilities into a separate file to remove a circular dependency.
```
==============================================
ProgramConstants.jl (`maxdegree, CONST_TYPE`)
Operators.jl
Equation.jl (`Node`)
Dataset.jl (`Dataset`)
    Options.jl
=============================================/ Core.jl
Core.jl
Utils.jl
        EquationUtils.jl
        EvaluateEquation.jl
            CheckConstraints.jl
            MutationFunctions.jl
            LossFunctions.jl
                PopMember.jl
                    Population.jl
                    HallOfFame.jl
                    ConstantOptimization.jl
        InterfaceSymbolicUtils.jl
            CustomSymbolicUtilsSimplification.jl
                SimplifyEquation.jl
                    Mutate.jl
                        RegularizedEvolution.jl
                            SingleIteration.jl
                                SymbolicRegression.jl <= Deprecates.jl, Configure.jl
```
So each file in this list can use `@show` on files to the left and above it. The `<=` just means I used `include` instead of `@from`.


patrick-kidger:
LGTM!
Regarding your questions about the two `include`s still existing: I think `include("Deprecates.jl")` should be replaced with `@from "Deprecates.jl" import RunSR, sqrtm, logm, logm2, logm10`. I'm less sure what's going on with `Configure.jl`. What kind of error do you get?

MilesCranmer:
Re: deprecation, this actually doesn't work with `@from`. This is because Base.@deprecate also exports the function from the module. But when I use `@from`, it doesn't get exported. So I think this is an edge case that still requires an `include(...)`. (cc @Roger-luo - probably of interest to you too)

MilesCranmer:
Re: `Configure.jl`, this is probably a less-common edge case (maybe this is the only package that experiences this!) so I wouldn't worry about it. I'll just use `include(...)`. But this is what I see as output:
```
Activating environment on workers.
      From worker 3:     Activating environment at `~/SymbolicRegression.jl/Project.toml`
      From worker 5:     Activating environment at `~/SymbolicRegression.jl/Project.toml`
      From worker 4:     Activating environment at `~/SymbolicRegression.jl/Project.toml`
      From worker 2:     Activating environment at `~/SymbolicRegression.jl/Project.toml`
Importing installed module on workers...Finished!
Testing module on workers...ERROR: LoadError: On worker 2:
UndefVarError: SymbolicRegression not defined
#9 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/macros.jl:87
#103 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/process_messages.jl:290
run_work_thunk at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/process_messages.jl:79
run_work_thunk at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/process_messages.jl:88
#96 at ./task.jl:356
Stacktrace:
 [1] #remotecall_fetch#143 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:394 [inlined]
 [2] remotecall_fetch(::Function, ::Distributed.Worker, ::Distributed.RRID) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:386
 [3] #remotecall_fetch#146 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:421 [inlined]
 [4] remotecall_fetch at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:421 [inlined]
 [5] call_on_owner at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:494 [inlined]
 [6] fetch(::Distributed.Future) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:533
 [7] test_module_on_workers(::Array{Int64,1}, ::Options{Tuple{typeof(+),typeof(*)},Tuple{typeof(cos)}}) at /mnt/home/mcranmer/SymbolicRegression.jl/src/Configure.jl:140
 [8] EquationSearch(::Array{Float32,2}, ::Array{Float32,1}; niterations::Int64, weights::Nothing, varMap::Nothing, options::Options{Tuple{typeof(+),typeof(*)},Tuple{typeof(cos)}}, numprocs::Nothing, procs::Nothing, runtests::Bool) at /mnt/home/mcranmer/SymbolicRegression.jl/src/SymbolicRegression.jl:179
 [9] top-level scope at /mnt/home/mcranmer/SymbolicRegression.jl/test/full.jl:29
 [10] include(::Function, ::Module, ::String) at ./Base.jl:380
 [11] include(::Module, ::String) at ./Base.jl:368
 [12] exec_options(::Base.JLOptions) at ./client.jl:296
 [13] _start() at ./client.jl:506
in expression starting at /mnt/home/mcranmer/SymbolicRegression.jl/test/full.jl:7
```

patrick-kidger:
> 
> 
> Re: deprecation, this actually doesn't work with `@from`. This is because `Base.@deprecate` also exports the function from the module. But when I use `@from`, it doesn't get exported. So I think this is an edge case that still requires an `include(...)`, or one to manually export the deprecated functions. (cc @Roger-luo - probably of interest to you too)

This at least is expected -- `@from` makes no attempt to export anything (since it's an importer, not an exporter!), so indeed I expect the correct thing to do is `@from "Deprecates.jl" import foo; export foo`. Sounds like we understand each other there.

The one for `Configure.jl` is altogether more mysterious. As we both know Distributed is very finickity, so I suppose this isn't a huge surprise.

I would note that FromFile operates by storing its compiled files in the rootmodule that it's called from. Meamwhile there's a difference between the master process and between the distributed workers, in that the first one is still in rootmodule `SymbolicRegression` at this point, but the workers all start off in `Main`. Theoretically there could be some funny business going on because of that. (Although I'd have hoped that the workers all just `@from` the files and compile+store them in their local `Main` without complaint.)

MilesCranmer:
Thanks! Okay, I'll leave that one for now. Most of the project is under FromFile.jl now, thanks again for the help! I think the imports are much cleaner now.

Cheers,
Miles

Issue #33, opened 2021-06-17
Saving equations throughout runtime 


cobac:
Hey, thank you so much for all your work in this package.

I'm interesting in evaluating the performance of this algorithm overtime (i.e. saving the `hallOfFame` object after each iteration). Is there any way that I've missed of achieving this without modifying the source code?

Thanks!

MilesCranmer:
Awesome to hear you are liking it! There is the `recorder::Bool` option in `Options`, which you can also turn on by setting the environment variable `PYSR_RECORDER=1`. Then, `recorder_file::String` is the filename for this (a JSON file; default is `pysr_recorder.json`). This JSON file records everything, even each individual mutation, which is maybe too much info, but it's helpful for debugging the performance. Does that work?

Cheers,
Miles

cobac:
Thanks! (btw those options are undocumented I don't know if you were aware)

Hmm... I assume that'd be quite a bit of overhead (?). But I could benchmark in different runs how the loss function evolves over time vs the actual time. Great.

However, I've done a few test runs with `recorder = true` (with and without `recorder_file = ...`) but I can't find the .json file in my filesystem. Using  [8254be44] SymbolicRegression v0.6.9. I can reproduce this in a clean environment.

Thanks again.

MilesCranmer:
No problem. That functionality is still experimental since I wrote it for myself to debug specific use-cases of the software in a project I'm using it on, but I can see now that others would find this useful too!

> However, I've done a few test runs with recorder = true (with and without recorder_file = ...) but I can't find the .json file in my filesystem. Using [8254be44] SymbolicRegression v0.6.9. I can reproduce this in a clean environment.

Sorry, I see the bug. Just pushed a fix to master. It should work now.

cobac:
Thanks!

I see that the program is trying to save the record now. However, all runs are failing at some point with `ERROR: Inf not allowed to be written in JSON spec`. Si I assume the program is trying to save an equation before checking if it produces proper results. I am not familiar enough with the code base to know what's the best workaround, but I assume you must have encountered the same issue in your own testing (?)

I've tested with current master and bc7b81ff80681b469bac55394c03ec56831e48e1

MilesCranmer:
I’ve never seen that before. How are you running it? Do you have a MWE?

cobac:
[...]

PR #36, opened 2021-06-22
Allow infinities in recorder

cobac:
The solution with checking the loss value for `Inf` https://github.com/MilesCranmer/SymbolicRegression.jl/commit/124020d79db1c766e6e35863b68bcc2608312a3d was failing, and I was testing with custom `println()` statements and couldn't find what was generating the `Inf`. But the JSON exporter allows to save Infinities with JS syntax.

Closes https://github.com/MilesCranmer/SymbolicRegression.jl/issues/33

### DIFF
--- a/src/Population.jl
+++ b/src/Population.jl
@@ -89,9 +89,8 @@ end
 
 
 function record_population(pop::Population{T}, options::Options)::RecordType where {T<:Real}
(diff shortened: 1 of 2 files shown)


MilesCranmer:
Very nice solution!! Thanks for this!

cobac:
Thanks to you!

PR #37, opened 2021-07-05
Updates for SymbolicUtils 0.13

AlCap23:
I've started writing an update for the symbolic conversion to make an interface with DataDriven.

I still need some time, but `node_to_symbolic` works now. All the other stuff will gradually happen over the next few days ;).


Project.toml
```diff
@@ -26,7 +26,6 @@ Optim = "0.19, 1.1"
 Pkg = "1"
 Reexport = "1"
 SpecialFunctions = "0.10.1, 1"
-SymbolicUtils = "0.6"
```
AlCap23, line 29: I forgot to include this again. But I should be working on the latest release.

AlCap23:
I had some issues with the explicit indexing of the functions, but this is resolved now. The tests are green on my machine and this seems to be ready to merge.

testfile.jl
```diff
@@ -0,0 +1,15 @@
+using SymbolicRegression
```
AlCap23, line 1: Ah, I added this without intention. Sorry. Will delete :)

MilesCranmer:
I am amazed; thanks so much @AlCap23! This looks **fantastic**. I will take a deeper look through this PR later this week to check through everything, but from an initial skim through, this looks really really good.

AlCap23:
I wonder if we need the `index_functions` keyword since I got rid of the naming convention. Maybe a tighter check via `typeof(f) === typeof(op)` is necessary.

src/InterfaceSymbolicUtils.jl
```diff
@@ -26,204 +26,204 @@ will generate a symbolic equation in SymbolicUtils.jl format.
...
-        if nargs == 2
-            r = symbolic_to_node(args[2], options, varMap=varMap)
-            return Node(op_idx, l, r)
-        else
-            # TODO: Assert operator is +, *
-            return _multiarg_split(op_idx, args, options, varMap)
-        end
+function findoperation(op, ops)
+    for (i,oi) in enumerate(ops)
+        Symbol(oi) == Symbol(op) && return i
```
AlCap23, line 89: Here.

AlCap23: Ah, forget it :) . I should really wait some time before waking up. We match the symbols.

MilesCranmer:
I changed `convert(::Node, eqn, options)` to be `convert(::Node, eqn, options; varMap=nothing)`. Is that okay with you? That allows the user to pass varMap for differently named variables, if necessary. (functionality preserved for `symbolic_to_node` too.).

MilesCranmer (commented):
Looks fantastic, thanks so much for pushing this! You have great software intuition; I was stuck for how to do this update, but this looks like a great way to do it.

Let me know if the small changes I added are good with you.

Final question before I merge: have you tested if the performance changes with the `try-catch` instead of explicitly returning `false` upon detecting inf or nan? I remember the `try-catch` hurting the equation evaluation significantly which is why I started doing explicit returns. Although, the simplification stuff isn't much of a bottleneck so it might not matter anyways.

Thanks again!
Miles

src/InterfaceSymbolicUtils.jl
```diff
@@ -5,225 +5,132 @@ using SymbolicUtils
...
+    using `symbolic_to_node`.
+"""
+function node_to_symbolic(tree::Node, options::Options;
+                     varMap::Union{Array{String, 1}, Nothing}=nothing,
+                     evaluate_functions::Bool=false,
+                     index_functions::Bool=false
+                     )
+    expr = subs_bad(parse_tree_to_eqs(tree, options, index_functions, evaluate_functions))
+    # Check for NaN and Inf
+    @assert isgood(expr) "The recovered equation contains NaN or Inf."
```
AlCap23, line 122: This is the reason for the try catch. We could just get rid of it and check the result for Inf / NaN here.

AlCap23:
Thanks! I am glad to hear this :).

I didn't benchmark the performance but included the try-catch due to the assertion error ( see my comment ) .

We could just get rid of the assertion or include a boolean flag to skip it and test the result for NaN/Inf.

src/InterfaceSymbolicUtils.jl
MilesCranmer: I think it's fine as is!

MilesCranmer:
Awesome. Let me do a quick benchmark to make sure nothing major changed, and I'll merge.

MilesCranmer:
Nice, it looks like this actually sped up the simplification functionality!

MilesCranmer:
Thanks again! Merged.

MilesCranmer:
*I mean 0.6.11 :)

PR #39, opened 2021-07-26
Autodiff for equations

kazewong:
Adding a functionality to compute the gradient of a tree respect to its input feature.

The main modifications are in EvaluateEquation.jl, and the most significant function added should be `evaldiffTreeArray`. Note that these modifications have not been incorporate in the main loop, but `evaldiffTreeArray` can be called given a tree.

PR todo's:

- [ ] Incorporate this feature into the main script.
- [ ] Write comment strings

Future todo's (for a different PR):

1. Optimize speed with fused operator.
2. Generalize to higher order derivatives
3. Generalize to derivatives respect to constants for backpropagation.


src/SymbolicRegression.jl
```diff
@@ -24,6 +24,17 @@ export Population,
     simplifyWithSymbolicUtils,
     combineOperators,
     genRandomTree,
+    mutateConstant,
```
MilesCranmer, line 27: Do these need to be exported? You can manually use them with:

```
using SymbolicRegression
...
SymbolicRegression.mutateConstant(...)
```

Exporting stuff is mainly for functions that a user would commonly use, rather than internal functions

kazewong: I tried not exporting those functions I added then use what you suggested, but it seems I cannot find them during runtime. I need access to all most of the low level functions since I am calling those functions in another directory, just like how one would call a library. Please feel free to remove these added exports if it doesn't match the package's general use cases.

MilesCranmer: If you are experimenting, then I would just leave it as is, but not commit the code. You can add changes selectively with `git add -p`. If code is inside the package it should be able to access them with `@from ...`.

MilesCranmer: Actually I think it should work, if you just call `@from ...` inside SymbolicRegression.jl. No need to export. You should be able to do `SymbolicRegression.mutateConstant`.

kazewong: Yeah it works, it just doesn't tab complete so I thought it is not there

src/Options.jl
```diff
@@ -353,6 +356,22 @@ function Options(;
...
+    diff_binary_operators = Any[]
+    diff_unary_operators = Any[]
+
+    for op in binary_operators
+        diff_op(x,y) = gradient(op,x,y)
+        push!(diff_binary_operators, diff_op)
+    end
+
+    for op in unary_operators
+       diff_op(x) = sum(gradient(op,x))
```
MilesCranmer, line 368: Why is this one summed?

MilesCranmer:
One other thing I forgot: I think diff_binops and diff_unaops should be added here as well, as a function set 4 and 5, respectively.

https://github.com/MilesCranmer/SymbolicRegression.jl/blob/ef319867a05ef7a6150c72760bd12e99d9d1698f/src/Configure.jl#L63-L74

This functionality moves user-defined functions or losses to other works (e.g., if using for multi-node). I think we'll have to move the derivatives as well, since Zygote is called outside of the workers.

MilesCranmer:
1. I modified your todo list to list things for the PR (before merge), and things for later. e.g., higher derivatives is probably too hard, we can leave for later.
2. I think a flag should be added to options for turning on/off the derivative calculation. For example, maybe the user defines a non-differentiable operator.
3. Optimizations (optional)
    1. I think we can speed up the derivative calculation a bit. Right now I think it does about O(n^n), for n the size of a tree, more computation than it needs to, since it re-runs evalTreeArray at each node. I think a single evalTreeArray can be called, and the result for each subtree can be stored for when computing the derivatives.
    2. I think a quick search could be run on each subtree before doing computation. If a feature does not appear in the subtree, the gradient is zero.

src/Options.jl
kazewong: It was a debug thing that I forgot to change it back. Tho I think it doesn't matter whether it is gradient(op,x)[1] or sum(gradient(op,x)) since there is only one component in the result

MilesCranmer:
I am still wondering if there is an easy way to do the gradient computation with a single pass through. e.g., the function `diff_deg0_eval` would return a vector for each row like `[0, 0, 0, 1, 0, 0]` where the `1` corresponds to if the 4th feature is in that node. Thoughts?

kazewong:
> I am still wondering if there is an easy way to do the gradient computation with a single pass through. e.g., the function `diff_deg0_eval` would return a vector for each row like `[0, 0, 0, 1, 0, 0]` where the `1` corresponds to if the 4th feature is in that node. Thoughts?

I just finish making that function and testing it in another branch. It agrees with mathematica for the cases I have tested.

MilesCranmer:
Any idea why the gradient test no longer passes?

kazewong:
> Any idea why the gradient test no longer passes?

I think in the test script, the line 

`@test evaldiffTreeArray(tree, input_data, options, i)[1][1] ≈ true_grad[i]`

is still getting the first argument from the function, which changed from the derivative to the values. 

`@test evaldiffTreeArray(tree, input_data, options, i)[2][1] ≈ true_grad[i]` should fix it

MilesCranmer:
Thanks! This looks fantastic. Reviewing now. I did some small changes in the code (like making them into for-loops, so I can add a check for infs/nans) but it should be equivalent. Questions:

- What are the contents of gradient_list? 
- Can you put a unit test into the test/derivative_test.jl for the grad function? I tried it on the equation ` tree = Node("x1") + cos(Node("x1")) * Node("x2") * 3.2` (which is equal to `(x1 + ((cos(x1) * x2) * 3.2))`). This seems to give the correct gradient for x, but wrong gradient for y. Not sure why this is.
- I made a separate function for `evalgradTreeArray` that simply has no `gradient_list` argument, so the user can call it without needing to pass an empty matrix themselves.
- Can the `constant_index` be moved into a separate tree type that is created for the gradient evaluation? i.e., instead of putting it into the `Node` type. Since it's only used for the gradient calculation, this keeps the `Node` type lighter weight, and also prevents future errors since the `constant_index` will become innaccurate after mutations (and needs to be recalculated anyways).
- You don't need to use `deepcopy` for arrays, `copy` gets everything that is needed and is much faster. `deepcopy` is for user-defined types generally. I'll put this change in too.

MilesCranmer:
Okay, I found the error! The evalgradTreeArray was doing binary operators correctly, but unary operators it was not yet batching over the `dcumulator`. I think this is fixed now.

In addition, I removed the `constant_index` from the `Node` struct, and instead implemented a `NodeIndex` struct, whose only purpose is to record indices for nodes in an equation. It's also a tree, but only has a single attribute for the index. I updated the `indexConstants` struct to output this tree, and the gradient operations to use this.

MilesCranmer:
Added a test for the gradient function, and also switched to CamelCase (`evalGradTreeArray` and `evalDiffTreeArray`).

MilesCranmer:
I'm a bit confused - the unittest passes on my own machine, but doesn't seem to pass in the GitHub action. What do you see on your machine?

(you can run it with `julia --project=.` then run `import Pkg; Pkg.test()`

kazewong:
> I'm a bit confused - the unittest passes on my own machine, but doesn't seem to pass in the GitHub action. What do you see on your machine?
> 
> (you can run it with `julia --project=.` then run `import Pkg; Pkg.test()`

Works fine for me on my workstation, which is running cent OS

MilesCranmer:
Hey @kazewong, sorry for the long delay on this. I spent today working on this, and finally resolved all the merge conflicts, incorporated the latest evaluation optimizations into the evalGrad/Diff functions, and expanded the unit tests. The unit tests now check both the gradients with respect to variables, and with respect to constants. I think this is now ready to go!

One last to-do item: I think this needs to check whether the derivative of an operator exists upon creating the differential operations. If the derivative does not exist, then it should store the identity function instead. Then, when autodiff is incorporated into the optimizer, then having non-differentiable functions should cause the optimizer to go back to its current strategy of numerically estimating the gradients.

src/Options.jl
```diff
@@ -318,6 +319,22 @@ function Options(;
     binary_operators = map(binopmap, binary_operators)
     unary_operators = map(unaopmap, unary_operators)
 
+    diff_binary_operators = Any[]
+    diff_unary_operators = Any[]
+
+    for op in binary_operators
+        diff_op(x, y) = gradient(op, x, y)
```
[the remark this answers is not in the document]
MilesCranmer: Okay added it. Now it displays the following message:

```julia
julia> options = Options(;
           binary_operators=(+, *, -, /, ^),
           unary_operators=(cos, exp, sin, myrelu),
       )
┌ Warning: Automatic differentiation has been turned off, since operator myrelu does not have well-defined gradients.
└ @ SymbolicRegression.var"../Options.jl" ~/Documents/SymbolicRegression.jl/src/Options.jl:362
```

where we have defined `myrelu` as `myrelu(x) = x > 0 ? x : 0` - which returns `nothing` from `Zygote.gradient` for negative input..

MilesCranmer (approved):

MilesCranmer (approved):

Issue #46, opened 2021-09-27
Version of SymbolicUtils.jl blocked to 0.13 while the current version is 0.16


hurak:
Installing `SymbolicRegression.jl` force downgrade of `ModelingToolkit.jl`, `Symbolics.jl` and `SymbolicUtils.jl`. I can read in `Project.toml` the compat entry `SymbolicUtils = "0.13"`. If I understand the convention described in [https://pkgdocs.julialang.org/v1/compatibility/#compat-pre-1.0](https://pkgdocs.julialang.org/v1/compatibility/#compat-pre-1.0) correctly, this binds the package to `0.13.x` versions. Is that necessary here? `SymbolicUtils.jl` is now in version `0.16.0`.

MilesCranmer:
See https://github.com/MilesCranmer/SymbolicRegression.jl/pull/45. Upgrading SymbolicUtils currently breaks the test suite, so will need to figure out why before adding the updated version.

MilesCranmer:
Actually, it looks like the tests pass. So it’s probably not necessary! But for stability reasons I need to still read the change logs before merging.

AlCap23:
#50 Seems fine on my machine.

MilesCranmer:
Version 0.6.15 should be up soon which satisfies this. Sorry for the delay.

PR #50, opened 2021-11-04
SymbolicUtils v0.18

AlCap23:
Updates for usage with `SymbolicUtils` from `v0.18`. 

I've adapted one simplification rule, which errored during precompilation and substituted `has_trig` with `has_trig_exp`.

Locally, the tests work except for one issue with an unsafe stream ( possible M1 related 😅 ).


MilesCranmer:
Awesome, thanks @AlCap23! Closes #51 #46.

PR #75, opened 2022-03-09
Refactoring PopMember + adding adaptive parsimony to tournament

MilesCranmer:
This PR includes both:

1. A refactor of PopMember to both include a score (which has the complexity penalty), and loss (the raw loss).
2. Incorporation of the adaptive parsimony strategy into the tournament scoring, rather than just the simulated annealing part of the pipeline.
3. Tweaking the adaptive parsimony to use a moving window (10 cycles window size), rather than accumulating over the entire search.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SymbolicRegression"
 uuid = "8254be44-1295-4e6a-a16d-46603ac705cb"
 authors = ["MilesCranmer <miles.cranmer@gmail.com>"]
-version = "0.7.14"
+version = "0.8.0"
 
 [deps]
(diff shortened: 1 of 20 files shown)


MilesCranmer:
This is a breaking change to the library as it introduces the `loss` attribute of `PopMember`, as well as changes `scoreFunc` to output two numbers rather than one. Therefore this will bump the version to 0.8.0.

MilesCranmer:
Generating random trees as one of the mutations kind of messes up the search because the trees are always size 5. Let me tweak the random tree generation mutation to be of random size - any possible size is valid.

Issue #59, opened 2022-01-24
Issues with simplification backend


MilesCranmer:
cc @AlCap23 

I've noticed many times during training that equations are overly complex, so I think simplification isn't working for some reason. I narrowed it down to this:
https://github.com/MilesCranmer/SymbolicRegression.jl/blob/561c9d7270073e28b1c82d43d7c328c5af0ebb1a/src/SimplifyEquation.jl#L153-L155

I think this is too general and is hiding the current issues in the simplifier. When I remove it, I see that many times, simplification is skipped for equations that should normally be easy to simplify.

@AlCap23 do you think you could help me fix the `convert` functions to capture the current API of SymbolicUtils? I think it will greatly improve the results of EquationSearch. Thanks!

MilesCranmer:
@AlCap23 this might be useful for debugging:

```julia
    catch e
        @warn "Error in symbolic_util simplify: $(e)\nFrom initial equation: $(stringTree(init_node, options))"
        return init_node
    end
```
where the preamble needs:
```julia
using Logging
@from "EquationUtils.jl" import countNodes, stringTree
```

MilesCranmer:
For example, the following tree:
```julia
> tree = Node("x1") * Node("x1") + Node("x1") * Node("x1") + 3f0
> printTree(tree, options)

(((x1 * x1) + (x1 * x1)) + 3.0)
```
cannot be simplified:
```julia
> simplifyWithSymbolicUtils(tree, options)
Error in symbolic_util simplify: ErrorException("Operation ^ in expression not found in operations (+, *, /, -)!")
```
SymbolicUtils normally converts multiplication into a polynomial basis, then collects common terms, which is fine, because at the end I run them through `multiply_powers` to convert `^` into repeated multiplication. However it looks like `multiply_powers` is not being called so the `^` is being left in the SymbolicUtils version of the equation, which results in an error.

`multiply_powers`:
https://github.com/MilesCranmer/SymbolicRegression.jl/blob/561c9d7270073e28b1c82d43d7c328c5af0ebb1a/src/CustomSymbolicUtilsSimplification.jl#L16-L78

MilesCranmer:
I see a problem:
```julia
julia> using SymbolicUtils
julia> @syms x
julia> x * x
x ^ 2
```
This must have become the default behaviour in a new SymbolicUtils. This is why simplification is broken.

MilesCranmer:
[...]

PR #84, opened 2022-04-08
Bump SymbolicUtils.jl to 0.19

ChrisRackauckas:
I'm having some trouble isolating the issue due to FromFile.jl eval. But it seems like it's:

```julia
using Metatheory
rhs = :(~(x^-1))
Metatheory.Syntax.makeconsequent(rhs)
[9:50 PM] julia> Metatheory.Syntax.makeconsequent(rhs)
ERROR: Error when parsing right hand side
Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:33
 [2] makeconsequent(expr::Expr)
   @ Metatheory.Syntax C:\Users\accou\.julia\dev\Metatheory\src\Syntax.jl:64
 [3] top-level scope
   @ REPL[2]:1
```

@shashi


MilesCranmer:
Thanks for the PR Chris. I think this PR also needs to revert this particular commit in master: https://github.com/MilesCranmer/SymbolicRegression.jl/commit/3b2608da21a0c9c57282cf40feac7649ef9849d9 to get the newer SymbolicUtils.jl working again, since the data types are quite different for `>0.6`. The other issue I think is with the simplification backend which we discussed on #59–we will need to automatically convert exponents into repeated multiplication when converting back to SymbolicRegression.jl.

MilesCranmer:
Another option may be to just re-implement the export functionality from scratch to exclusively do exports and nothing else? This would be much less work I think.

I have the SymbolicUtils.jl simplification utilities turned off by default anyways since I noticed they were taking an exponentially longer amount of time for very complex expressions, so we could simply remove it entirely for now, and just leave a simple export functionality. The simplification functionality was a bit overkill anyways to have within a genetic algorithm since it will try to simplify the entire expression every single time called, when in reality I would want something a bit more stochastic/lightweight to partially simplify only part of an expression every time it is called.

MilesCranmer:
@ChrisRackauckas FYI I switched to a normal `include()`-based system so it should be easier for you to use normal code analysis tools to find this.

ChrisRackauckas:
@shashi wasn't something like `@variables x::SafeFloat` added for removing some of the simplifications? That might be what's needed here.

MilesCranmer:
I think it's easier if we leave out the simplification for now and add it back in later. For the simplification to work i realized I need to experiment with some sort of "max_evals" to the simplifier, so that very complex expressions don't slow down the overall search too much.

The edits I pushed yesterday are basically removing the simplification parts, and simply leaving in the import/export utils, so it can interface with SymbolicUtils.

shashi:
`@variables x::LiteralReal`, there is also `SafeReal` which is to avoid cancellation in division only.

MilesCranmer:
Seems like it's working now which is good. I'm going to try to get 100% test coverage on the SymbolicUtils interface before merging, just in case.

src/InterfaceSymbolicUtils.jl
```diff
@@ -4,12 +4,108 @@ using SymbolicUtils
...
+const SUPPORTED_OPS = (cos, sin, exp, cot, tan, csc, sec, +, -, *, /)
+
+isgood(x::SymbolicUtils.Symbolic) = SymbolicUtils.istree(x) ? all(isgood.([SymbolicUtils.operation(x);SymbolicUtils.arguments(x)])) : true
+subs_bad(x) = isgood(x) ? x : Inf
+
+function parse_tree_to_eqs(tree::Node, options::Options, index_functions::Bool=false)
+    if tree.degree == 0
+        # Return constant if needed
+        tree.constant && return subs_bad(tree.val)
+        return SymbolicUtils.Sym{Number}(Symbol("x$(tree.feature)"))
```
MilesCranmer, line 18: This one should be `LiteralReal` instead of `Number`. Same for other instances.

MilesCranmer:
Cool, `LiteralReal` seems to work - thanks @shashi!

```julia
using SymbolicUtils, SymbolicRegression
options = Options(binary_operators=(+, *))
tree = Node("x1") * Node("x1")
eqn = convert(Symbolic, tree, options)
tree_copy = convert(Node, eqn, options)
println(tree_copy)  # =(x1 * x1)
```
Before, this would give an error as it tries to convert it to `x1^2`, while `^` is not specified as an operator. But now it works and doesn't do simplification.

MilesCranmer (approved):

MilesCranmer:
Thanks for starting this PR, happy to have this finally working!

MilesCranmer:
See https://github.com/SciML/DataDrivenDiffEq.jl/pull/346 for the update to DataDrivenDiffEq.jl

ChrisRackauckas:
Awesome, thanks for completing this.

PR #86, opened 2022-04-10
[WIP] Optimize expressions with Zygote-based autodiff

MilesCranmer:
cc @kazewong 

This is an attempt to use `evalGradTreeArray` in the constant optimization. This function computes the explicit Zygote.jl-based derivatives of an expression with respect to its constants, so _in theory_ it should be faster than letting Optim.jl figure the gradient out using finite differences.

However, right now this attempt is about a quarter of the speed of when using `Options(..., enable_autodiff=false)`. I'm not sure why this is. @kazewong, you mentioned you had seen a speedup right? Do you know why this attempt is slowing things down?

Cheers,
Miles


MilesCranmer:
I also messed around with using `Optim.only_fg!` so I wouldn't have to re-compute the expression output. I couldn't get this warning as it seemed to want the Hessians too... Attempt is on branch `optimize_with_autodiff_shared_compute`.

MilesCranmer:
Alternatively... maybe it only appears to be slower since the optimization is limited by number of steps? i.e., it only appears slower since I am now computing, say, 10 evaluations of f + 10 evaluations of df, vs before where it might have been 20 evaluations of f. The former should give you **better** optimization, even though it's slower. So perhaps a better comparison would be with a reduced number of optimization steps? (a bit harder to measure - will need to explicitly measure the constant optimization routine with/without autodiff)

MilesCranmer:
So many changes since this was submitted we probably want to re-write this from scratch.

PR #91, opened 2022-04-18
fix worker connection timeout error

CharFox1:
**The bug:**  Workers will sometimes fail to connect to each other after initialization.  If a task for a worker is longer than the default connection timeout (60s) and another worker tries to connect to the first, they will fail.  

**The fix:**  Setting `lazy=false` forces all connections to be created when workers are, meaning they do not need to be created while things are running and the above error cannot happen.  While this does add some work to startup, added computation time is negligible unless running with a very large number of workers.


MilesCranmer:
Thanks for the awesome contribution @CharFox1!! This is great. I never realized where these disconnects were coming from.

PR #100, opened 2022-05-02
Apply Blue formatting + change all internal methods to snake_case

MilesCranmer:
This applies the "blue style" Julia formatting to every single file in the repository. This will cause changes for branched code, so to fix this, it is probably easiest if you apply the same formatting to the branch, then manually git patch things.

The formatting can be done with:
```julia
using JuliaFormatter
format(".")
```
in the root directory.

The other changes were a bulk regex replace of the deprecations listed in `src/Deprecations.jl` using the Python regex script in `src/scripts/apply_deprecates.py`, with manual corrections where necessary (e.g., `loss` is often used as a variable, but it is now a function in the library).

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,1 @@
+style = "blue"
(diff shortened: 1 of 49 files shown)


MilesCranmer:
Last change is to add a GitHub action to automatically run this on pull requests and periodically on the main branch.

PR #116, opened 2022-08-20
Set `timeout-minutes` for CI

rikhuijzer:
Most jobs complete in about 20 minutes, so anything above 60 minutes can probably be considered a failure. This avoids having to wait 6 hours for a failure to be reported.


MilesCranmer:
Thanks!

PR #119, opened 2022-08-30
Generalize expressions to have arbitrary constant types

MilesCranmer:
This rewrites much of the equation interface to allow for arbitrary types for constants in each expression. Right now it is just `AbstractFloat`, but in principle it could be extended to allow for complex constants in expressions.

There seem to be some issues in computing directional derivatives of expressions, but everything else seems to work well. Surprisingly, it's even faster than before. I think this is because there are fewer conversion steps and hence more type stability; the equation evaluation can safely assume the type of data and the type of constants is the same.

cc @ChrisRackauckas in case of interest.

---

To use this, simply pass an array with the same precision as you desire having for the constants. Passing data of `Array{Float64}` means you get trees of type `Node{Float64}` back. Before this change, constants were all fixed to be `Float32`.


MilesCranmer:
Still need to check whether this impacts overall search performance... Can check this with the deterministic mode.

MilesCranmer:
Doesn't seem to affect performance.

MilesCranmer:
@kazewong want to have a look? This tweak might be good for expressions that have large dynamic ranges - you can now use `Float64` constants (and even `BigFloat` - arbitrary precision - though requires a manual addition of `randn`).

MilesCranmer:
@ChrisRackauckas in terms of interfacing with other packages, do you think `AbstractFloat` generic enough, or should I use another type?

ChrisRackauckas:
`AbstractFloat` is probably best unless there's a nice way to support integers (/ rationals)

MilesCranmer:
Very interesting idea. In principle that should be possible, and it would let you find equations that have integral constants. I would just need to permit different type parameter for `Node` types in a few functions, and also add a few conversion steps.

Would `Optim.optimize` work for integers and rational constants? If not I could look at disabling it if a type is not set up for it.

ChrisRackauckas:
> Would Optim.optimize work for integers and rational constants? If not I could look at disabling it if a type is not set up for it.

Nope. Mixed integer programming is a whole different beast from what Optim wants to solve. In theory you could use JuMP there though.

MilesCranmer:
Thanks! It would be very interesting to try to integrate JuMP. Not just for constant optimization, but perhaps one could use SymbolicRegression.jl as a solver to find symbolic forms for problems specified using JuMP's syntax... that would give us a nice way to include constraints on the symbolic form.

I'll take your suggestion and keep this as `AbstractFloat` for now, then could look at generalizing in the future.

PR #122, opened 2022-09-06
Generalize node type to non-floats

MilesCranmer:
I had a request from someone over email that I enable SymbolicRegression.jl to work with FixedPointNumbers.jl since they wish to find expressions that work with this datatype for their specific hardware.

This change generalizes the `Node` type to constants that are any `Real`, rather than just `AbstractFloat`. To make this change I had to change the constructor from `Node(val::AbstractFloat)` and `Node(feature::Int)` to:

```julia
Node(; val::T1=nothing, feature::T2=nothing)
Node(::Type{T}; val::T1=nothing, feature::T2=nothing)
```

In other words, now you explicitly specify whether the node is a constant or a variable, like `Node(; val=2.0)` or `Node(; feature=1)`. This also lets you construct expressions which have integers for constants, although this won't work for the actual search, as `Optim` is not enabled for it.

This change lets you use `FixedPointNumbers` like so:
```julia
using SymbolicRegression
using FixedPointNumbers
options = Options(binary_operators=(+, *, /, -), unary_operators=(cos, sin))

# Set up some variable nodes:
x1, x2 = Node("x1"), Node("x2")

# By default, these will have type `Float32`:
println(typeof(x1))
# > Node{Float32}

# Create an expression. This will automatically
# promote types:
tree = cos(x1) * 3.2 - sin(5.2 / x2)
println(typeof(tree))
# > Node{Float64}

# Convert the expression to FixedPointNumbers:
T = Fixed{Int32, 8}
fixed_point_tree = convert(Node{T}, tree)

println(fixed_point_tree)
# > ((cos(x1) * 3.199Q23f8) - sin(5.199Q23f8 / x2))
```


MilesCranmer:
Probably want to include tests for custom `Real` types.

MilesCranmer:
This is a breaking change to the `Node` constructor, so this will be version 0.12 of SymbolicRegression.jl

Issue #109, opened 2022-06-05
[Feature] Use NaNMath instead of `abs`


MilesCranmer:
Operators should be default use NaNMath https://github.com/mlubin/NaNMath.jl, instead of my versions which simply map input to the valid domain like `log(abs(x))`. Since the evaluator already detects NaNs by default, this should work well and expressions will automatically try to avoid having improper inputs.

johanbluecreek:
I can take a look at this. I'm not too familiar with the whole code base yet, but right now it seems to more or less to trace the use of the operations defined in `src/Operators.jl` and replace the occurrences with `NaNMath.log` and so on. If you have any other thoughts on the approach before I start let me know.

I also wanted to ask, I just did a quick naive benchmark on `NaNMath` versus an own implementation and `NaNMath` seems to perform consistently a bit worse,

```julia
julia> using NaNMath

julia> function log_nan(x)
           x < 0.0 && return NaN
           return log(x)
       end
log_nan (generic function with 1 method)

julia> X = rand(50_000_000) .* 2 .- 1;

julia> @time NaNMath.log.(X);
  0.661889 seconds (132.60 k allocations: 388.716 MiB, 0.80% gc time, 6.57% compilation time)

julia> @time NaNMath.log.(X);
  0.685413 seconds (4 allocations: 381.470 MiB)

julia> @time NaNMath.log.(X);
  0.614610 seconds (4 allocations: 381.470 MiB)

julia> @time log_nan.(X);
  0.644774 seconds (117.23 k allocations: 387.860 MiB, 12.58% gc time, 11.42% compilation time)

julia> @time log_nan.(X);
  0.543851 seconds (4 allocations: 381.470 MiB)

julia> @time log_nan.(X);
  0.477451 seconds (4 allocations: 381.470 MiB, 3.61% gc time)
```

Comparing the two best times above, `NaNMath` performs some 29% slower. Maybe this is not something significant enough (or maybe I'm too naive in trusting this benchmark), and maybe it is easier to just go with `NaNMath` anyway, I just wanted to ask it that is still the desired solution.

MilesCranmer:
Also, after defining new operators with custom names like `log_nan`, you could map their function name to simply `log` when printed, by modifying https://github.com/MilesCranmer/SymbolicRegression.jl/blob/29f6bf19d0920f74fb79431418d3e34ea2c4af75/src/Equation.jl#L204

MilesCranmer:
[...]

PR #123, opened 2022-09-07
Use functions returning NaN on branch cuts instead of abs (issue #109)

johanbluecreek:
This PR is intended to resolve issue #109 . It replaces logarithmic `_abs`, `acosh_abs`, and `sqrt_abs` with corresponding functions returning `NaN` on their corresponding branch cuts. This also removes all regularizing terms in the arguments of the previous functions, e.g. the `1//100000000`. 

The corresponding functions for generic types are now mapped to non-regularized functions as well. As I understand the need for these function definitions is to map them to symbolic expressions for the package `SymbolicUtils.jl`. The symbolic expression the user gets may have branch cuts, but I think it should be up to the user to analyse those properties of the final expression instead of having the functions mapped to regularized versions of the same functions. But let me know if I'm missing some reason why the symbolic expressions should be regularized. There appear to be no unit-tests for this, let me know if some should be added.

The old tests have also been updated: the `test_tree_construction` tests are for the new functions only tested in their valid domain, and `test_operators` test points inside and outside the domain of the new functions.

To note, since `pow_abs` is marked with `@deprecate` I did not bother to change this operation. Let me know if there is something missing that should be fixed in relation to power expressions.

All tests pass on my computer, and I also did some searches that finished without any apparent errors.


src/Operators.jl
```diff
@@ -67,15 +76,13 @@ sub(x, y) = x - y
 mult(x, y) = x * y
 pow_abs(x, y) = abs(x)^y
 div(x, y) = x / y
-log_abs(x) = log(abs(x) + 1//100000000)
-log2_abs(x) = log2(abs(x) + 1//100000000)
-log10_abs(x) = log10(abs(x) + 1//100000000)
-log1p_abs(x) = log(abs(x) + 1)
-acosh_abs(x) = acosh(abs(x) + 1)
+log_nan(x) = log(x)
```
MilesCranmer, line 79: Could these be made equivalent to their non-generic versions - e.g., `(x <= 0) ? NaN : log(x)`?

src/Core.jl
```diff
@@ -30,12 +30,12 @@ import .OperatorsModule:
     pow,
     pow_abs,
     div,
-    log_abs,
-    log2_abs,
-    log10_abs,
-    log1p_abs,
-    sqrt_abs,
-    acosh_abs,
+    log_nan,
```
MilesCranmer, line 33: Could these be renamed to, e.g., `safe_log` instead of `log_nan`? What do you think?

src/Equation.jl
```diff
@@ -250,7 +250,12 @@ function string_tree(
             end
         end
     elseif tree.degree == 1
-        return "$(options.unaops[tree.op])($(string_tree(tree.l, options, bracketed=true, varMap=varMap)))"
+        if occursin("_nan", string(options.unaops[tree.op]))
```
MilesCranmer, line 253: It might be good to add a test for this function, so that `log_nan` is converted to, e.g., `"log"`.

Also I'm not sure if it should be this general, as a user could define their own function using `"_nan"` in the title - we wouldn't want to convert that. Maybe this could be a dictionary mapping for the internal operators?

src/Operators.jl
```diff
@@ -67,15 +76,13 @@ sub(x, y) = x - y
 mult(x, y) = x * y
 pow_abs(x, y) = abs(x)^y
```
MilesCranmer, line 77: `pow_abs` is another one I think should be converted, so that it returns NaN if `x <= 0`.

MilesCranmer (changes requested):
Thanks, this is looking good - left some comments. Let me know what you think.

src/Core.jl
MilesCranmer: (It might be easiest to do a regex replace for this, to save you some time)

src/Operators.jl
johanbluecreek: Not every `x` works in that context. If `x`, I presume, is `<: Number` it works, but for example in the context of `SymbolicUtils` `x` is not, and `x <= 0` cannot be evaluated. In other terms, if `x` is a symbolic expression, we do not know if `x` is smaller or equal to zero or not before it has been evaluated.

src/Core.jl
johanbluecreek: Sure, I can rename them.

src/Equation.jl
johanbluecreek: Sure, I can use a dictionary here instead. I sort of just assumed that `_nan` would be a rare enough name, but I agree.

src/Operators.jl
johanbluecreek: Oh, I misunderstood what `@deprecate` did. I will change this one too.

src/Operators.jl
MilesCranmer: Good point! Sounds good to me.

MilesCranmer:
Thanks, will take a look! Yes - you hit "Resolve Conversation" when a suggestion is addressed. I already clicked them though.

src/Operators.jl
```diff
@@ -37,26 +37,41 @@ end
 function cube(x::T)::T where {T<:Real}
     return x^3
 end
-function pow_abs(x::T, y::T)::T where {T<:Real}
-    return abs(x)^y
+function safe_pow(x::T, y::T)::T where {T<:Real}
+    if isinteger(y)
+        y < T(0) && x == T(0) && return T(NaN)
+    else
+        y > T(0) && x < T(0) && return T(NaN)
```
MilesCranmer, line 44: Shouldn't these be `||` for the check on `x, y`?

test/test_operators.jl
```diff
@@ -26,17 +27,27 @@ types_to_test = [Float16, Float32, Float64, BigFloat]
...
     @test mult(val, val2) == val * val2
     @test plus(val, val2) == val + val2
     @test sub(val, val2) == val - val2
     @test square(val) == val * val
     @test cube(val) == val * val * val
+    @test isnan(safe_pow(T(0.0), -T(1.0)))
+    @test isnan(safe_pow(-val, val2))
+    @test all(isnan.([safe_pow(-val, -val2), safe_pow(T(0.0), -val2)]))
+    @test abs(safe_pow(val, val2) - val^val2) < 1e-6
+    @test abs(safe_pow(val, -val2) - val^(-val2)) < 1e-6
```
MilesCranmer, line 50: Might be good to test that the safe operations give `NaN` when expected?

src/Operators.jl
```diff
@@ -37,26 +37,41 @@ end
 function cube(x::T)::T where {T<:Real}
     return x^3
 end
-function pow_abs(x::T, y::T)::T where {T<:Real}
-    return abs(x)^y
+function safe_pow(x::T, y::T)::T where {T<:Real}
+    if isinteger(y)
```
MilesCranmer, line 41: Would it be faster if this was `T <: Integer`?

Also - I'm not sure if `T(NaN)` will work if it's an integer.

src/Operators.jl
MilesCranmer: Wait, nevermind. This is correct. Let me resolve this...

src/Operators.jl
MilesCranmer: Actually I think there's just not any good way to have safe operations for integers. I think this piece of the code could be removed for the time being. We can come back to such types later, but not needed for this PR.

MilesCranmer: Wait, I misunderstood what `isinteger` actually does. This looks good then!

test/test_operators.jl
MilesCranmer: Actually looks like this is already there! Nevermind.

src/Options.jl
```diff
@@ -127,11 +127,11 @@ The current arguments have been tuned using the median values from
...
 # Arguments
-- `binary_operators`: Tuple of binary
-    operators to use. Each operator should be defined for two input scalars,
-    and one output scalar. All operators need to be defined over the entire
-    real line (excluding infinity - these are stopped before they are input).
-    Thus, `log` should be replaced with `log_abs`, etc.
+- `binary_operators`: Tuple of binary operators to use. Each operator should
+    be defined for two input scalars, and one output scalar. All operators
+    need to be defined over the entire real line (excluding infinity - these
+    are stopped before they are input), or return `NaN` where not defined.
```
MilesCranmer, line 133: Nice, thanks for adding this!

test/test_print.jl
```diff
@@ -31,3 +31,19 @@ EquationSearch(
 s = repr(tree)
 true_s = "((sin(cos(sin(cos(v1) * v3) * 3.0) * -0.5) + 2.0) * 5.0)"
 @test s == true_s
+
+for unaop in [safe_log, safe_log2, safe_log10, safe_log1p, safe_sqrt, safe_acosh]
```
MilesCranmer, line 35: Nice, thanks for adding!

MilesCranmer (approved):
Looks great, thanks for the contribution!

PR #132, opened 2022-09-14
Use `LoopVectorization.@turbo` for evaluation loops

MilesCranmer:
This switches from `@inbounds @simd` to `@turbo` for most of the inner evaluation loops. There is a 30% speedup which is huge.

However, there's currently a `StackOverflowError` when running the `Float16` tests. I'm also not sure if LoopVectorization will be safe to use, since ideally I want to allow for **any** user-defined operator - would it mean certain passed operators fail? Also need to test it on a distributed system.


MilesCranmer:
Okay looks like LoopVectorization.jl hasn't even implemented `SpecialFunctions`, and will raise a StackOverflowError if any of them are used (as an operator): https://github.com/JuliaSIMD/LoopVectorization.jl/issues/233. I'm only seeing it in `Float16` since that's the first test (which tries a `gamma` function as an operator).

So this change seems not possible at the moment, unless `@turbo` is capable of falling back to non-SIMD operations to avoid StackOverflowErrors.

MilesCranmer:
Raised issue on https://github.com/JuliaSIMD/LoopVectorization.jl/issues/430

MilesCranmer:
Probably also want to switch to using `similar` throughout, rather than `Array{T,1}(undef, n)`. This is also needed independent of this change for generality to input array type.

MilesCranmer:
With https://github.com/JuliaSIMD/LoopVectorization.jl/pull/431 being merged, this should now work. Let's see how stable things are.

PR #136, opened 2022-10-05
Refactor parts of main equation search

MilesCranmer:
This PR greatly refactors parts of the main equation search, making it much more readable and generic. 

1. It creates a new file `SearchUtils.jl` to hold functions for code snippets used throughout the main search loop.
2. It moves the baseline loss to `Dataset`, rather than needing to pass a separate baseline loss to functions. However, this requires one to make the dataset object mutable, so that the equation search can set the baseline loss at runtime.
3. Moves the adaptive parsimony into a `RunningSearchStatistics` struct inside a new file `AdaptiveParsimony.jl`
4. Moves all early stop code into functions inside `SearchUtils.jl`.


MilesCranmer:
Tests:

- [x] Smoke test with same random seed: should give exactly the same result.
- [ ] Distributed multi-node test. Nothing should break; should have good performance.

src/ProgressBars.jl
```diff
@@ -263,6 +263,16 @@ function newline_to_spaces(string, terminal_width)
...
+mutable struct WrappedProgressBar
+    bar::ProgressBar
+    state::Union{Int,Nothing}
+    cycle::Union{Int,Nothing}
+
+    function WrappedProgressBar(args...; kwargs...)
+        return new(ProgressBar(args...; kwargs...), nothing, nothing)
+    end
+end
 # end
```
MilesCranmer, line 277: Probably want to put some kind of `update_progress!` here, to avoid having to update it manually in `SearchUtils.jl`.

PR #142, opened 2022-10-13
Attempt to boost startup time with SnoopPrecompile

MilesCranmer:
@Quarlos this demonstrates a simple implementation of SnoopPrecompile. Using this for precompilation (i.e., `Pkg.precompile()`) reduces startup time for me by about **33%**. Want to have a look at this and see if you can improve it at all?

Here's the code I tested this on:

```julia
@time begin
    import Random: seed!
    using SymbolicRegression
    seed!(0)
    X = randn(Float64, 2, 3); y = X[1, :];
    options = Options(progress=false, unary_operators=(cos,), max_evals=10, verbosity=0)
    hof = EquationSearch(X, y, options=options, multithreading=true);
    nothing
 end
```

~~The one strange thing is that it seems to complain about me redefining operators on the `Node{T}` type compared to the precompiled version:~~
```julia
WARNING: Method definition +(T1, SymbolicRegression.CoreModule.EquationModule.Node{T2}) where {T1<:Real, T2<:Real} in module OptionsModule at /Users/mcranmer/Documents/SymbolicRegression.jl/src/Options.jl:560 overwritten on the same line (check for duplicate calls to `include`).
  ** incremental compilation may be fatally broken for this module **
```
~~I do this redefinition because the `Node{T}` uses an enum over operators, so for things like `print` to actually be helpful, I have to redefine them every time an `Options` is created. So, for using this in practice, it would probably be a good idea to prevent SnoopCompile from precompiling those functions.~~

Fixed the above issue with an extra parameter that disables it!

cc @ChrisRackauckas - thanks for the recommendation.


MilesCranmer:
Probably the simplest way to fix these precompile warnings is to add a flag to `Options` that simply tells it to not re-define the base functions.

MilesCranmer:
New version in #161

PR #143, opened 2022-10-14
Fully-customizable training objectives

MilesCranmer:
Right now the only way to define a custom loss function is with an elementwise loss. You could hack `eval_loss` manually, but this is not very modular.

This PR (still a draft) makes it so that the _entirety_ of `eval_loss` can be defined by the user.

### This means you can basically symbolically optimize any objective or constraint you want!

You could write down `eval_loss` to incorporate derivative information (via `eval_grad_tree_array`), or do some correlation metrics on the predicted datasets, do numerical integration, or even work with PDE-like data.

Here's an example:

```julia
function my_objective(tree, dataset::Dataset{T}, options)::T where T
    (prediction, completion) = eval_tree_array(tree, dataset.X, options)
    !completion && return T(Inf)  # (In case of NaNs)

    # multiply the dataset by 2, and reevaluate?
    (prediction2, completion) = eval_tree_array(tree, dataset.X .* 2, options)
    !completion && return T(Inf)

    # evaluate the gradient with respect to features, with X -> 3X?
    (prediction3, gradients, completion) = eval_grad_tree_array(tree, dataset.X .* 3, options; variable=true)
    !completion && return T(Inf)

    truth = dataset.y
    
    return my_crazy_loss_function(truth, prediction, prediction2, prediction3, gradients)
end

options = Options(; loss_function=my_objective)
EquationSearch(X, y; options=options, multithreading=true)
```

---

cc'ing in case of interest @kazewong @CharFox1 @johanbluecreek @ChrisRackauckas @pitmonticone @AlCap23 @Patrick-Kidger

Any thoughts on this API/things to improve?


johanbluecreek:
Maybe this points are discussions more appropriate somewhere else, but I have some questions in regard to solving differential equations, PDE, as you mention, as a special case:

* How would e.g. boundary conditions be included in your loss function? I guess you would not specify that as an input into the search, instead it is just an addition to the loss function one would define, or?
* The usual DE solving trick of reducing differential degrees by including more functions, e.g. $\phi'' \phi' + 4/x^3 = 0$ is solved by a system { $\phi' = \psi$, $\psi ' \psi + 4/x^3$} instead. I guess that is what is imagined you'd also do with `SymbolicRegression.jl`? But that is really hard for the solver to do, right? Instead of just having to find the expression $\phi$ via genetic operations it has to find two expressions $\phi$ and $\psi$ that are just separated by *only* a differentiation operation. One additional differentiation operation at each evaluation of the loss should be much less costly. As I understand the differentiation operation `eval_grad_tree_array` the differentiated tree is not returned and higher order differentials are not available to the user. Would that not be useful to include at some point?
* Solving e.g. a DE you may just be looking at an interval, say, solve $\phi'' \phi' + 4/x^3 = 0$ on $x \in [1,10]$. The whole $\phi'' \phi' + 4/x^3$ would be part of the user's loss function, rendering the `y` input unnecessary. It would be nice to make this optional for such cases, otherwise it may be confusing for the user?

Maybe these issues are covered somewhere else, then I apologize for being a bit off topic. If you think any of these points are worth considering, but not here. Let me know if I should move it to an issue or something then.

MilesCranmer:
> * The usual DE solving trick of reducing differential degrees by including more functions, e.g. ϕ″ϕ′+4/x3=0 is solved by a system { ϕ′=ψ, ψ′ψ+4/x3} instead. I guess that is what is imagined you'd also do with `SymbolicRegression.jl`? But that is really hard for the solver to do, right? Instead of just having to find the expression ϕ via genetic operations it has to find two expressions ϕ and ψ that are just separated by _only_ a differentiation operation. One additional differentiation operation at each evaluation of the loss should be much less costly. 

Interesting. One easy way to get multiple expressions out is to enforce that the root node is a binary operator (or else, an infinite loss is returned), and is ignored. Then, `tree.l` is ϕ and `tree.r` is ψ, and you can do your evaluation on each, using separate calls to `eval_tree_array`. Sometimes it might delete the right branch, but that's okay since those branches of the evolutionary tree would just die out, given their large loss.

> As I understand the differentiation operation `eval_grad_tree_array` the differentiated tree is not returned and higher order differentials are not available to the user. Would that not be useful to include at some point?

It's definitely possible; it's just an engineering/effort challenge at this point. The hard part is because we assume a particular set of operators, we have to compile-in all the first-order derivatives of every operator (with `Zygote.jl`, and saved to `Options`). So for higher-order differentiation, you would need to compile-in the higher-order derivatives of each operator.

> * Solving e.g. a DE you may just be looking at an interval, say, solve ϕ″ϕ′+4/x3=0 on x∈[1,10]. The whole ϕ″ϕ′+4/x3 would be (or perhaps rather _could_ be) part of the user's loss function, rendering the `y` input unnecessary. It would be nice to make this optional for such cases, otherwise it may be confusing for the user?

Good point! Yes, it would probably be good to make `y` optional. And perhaps also make it so that you could pass additional data into `Dataset`, like 3D arrays and other types of data.

> * How would e.g. boundary conditions be included in your loss function? I guess you would not specify that as an input into the search, instead it is just an addition to the loss function one would define, or?

(Moved this down from above.) Maybe this could be a part where it would be useful to allow other types of data in `Dataset`.

MilesCranmer:
With custom losses, I should also get rid of the requirement for losses to be non-negative. It’s really only for the scoring part, but I could just compute the scores with losses subtracted by the min loss; it’s no different.

MilesCranmer:
cc @astrockragh

PR #147, opened 2022-10-21
Split codebase in two: DynamicExpressions.jl and SymbolicRegression.jl

MilesCranmer:
This splits the entire codebase in two:
- SymbolicRegression.jl (this repo) will now focus on the symbolic regression stuff exclusively.
- [DynamicExpressions.jl](https://github.com/SymbolicML/DynamicExpressions.jl) now exposes the core data structures used in this repository: the `Node` type, and the evaluation kernels.

This reduces the code complexity quite a bit, and makes it easier to focus on the actual evolutionary algorithm in one codebase, and an efficient expression evaluation scheme in the other codebase. Already by porting DynamicExpressions.jl into its own codebase, I've found it's easier to get tensor-based evaluation up-and-running! See the README example for one where I even define an expression on strings and then evaluate it: https://github.com/SymbolicML/DynamicExpressions.jl.

In the future it would be great if SymbolicRegression.jl was completely generic to type. A good test scenario of this would be: can I optimize a **string**, using a set of operators on strings, to have a particular loss metric?

This will then make it super easy to get a tensor version of symbolic regression working.

TODO:

- [x] Override for `expression(X)`.
- [x] Remove unused modules (e.g., I think `LinearAlgebra` is no longer imported here)


src/Options.jl
```diff
@@ -326,6 +325,9 @@ function Options(;
     enable_autodiff::Bool=false,
     nested_constraints=nothing,
     deterministic=false,
+    # Not search options; just construction options:
+    extend_user_operators=true,
```
MilesCranmer, line 329: Needs docstring

MilesCranmer:
Seems like it fails a lot for the state saving test, and might even segfault. I guess there’s some state that’s not being initialized correctly?

PR #154, opened 2022-11-02
Do optimizations as a low-probability mutation

MilesCranmer:
Right now, optimization is performed after an iteration for all population members. However, this seems suboptimal for cases where `ncyclesperiteration` is large - so this introduces optimization as a new type of mutation operator. Optimization is expensive so this is very low probability by default.

This should be useful for distributed searches, where you want to have limited communication with the head node - and therefore want to be performing constant optimization as part of the mutations.


MilesCranmer:
Probably want to leave this as 0.0 probability until the implications are understood.

PR #161, opened 2022-11-27
Proper SnoopCompilation

MilesCranmer:
This does a few things which give vastly better startup times:

1. `SnoopPrecompile.@precompile_all_calls` used to precompile all relevant functions.\
2. DynamicExpressions.jl is also using SnoopCompile for better pre-compilation as of 0.4.2, and startup is pretty much instantaneous (aside from `using`). See https://github.com/SymbolicML/DynamicExpressions.jl/pull/11.
3. DynamicExpressions.jl now does not specialize to the order of operators, only individual functions, so all common operators can now be pre-compiled.
4. Most type specialization from `Options` has been removed, where it did not improve the speed. The only remaining type specialization is the numeric type for complexity

The time of `using SymbolicRegression` is now the most expensive, so it's worth it to build a sysimage with the following snippet:

```bash
julia -O3 --threads=auto --startup-file="no" -e 'using Pkg; Pkg.activate(;temp=true); pkg"add SymbolicRegression.jl#snoop-compile2"; pkg"add DynamicExpressions"; pkg"add PackageCompiler"; open("precompile_file.jl", "w") do io; write(io, "using DynamicExpressions; using SymbolicRegression"); end; using PackageCompiler; create_sysimage(["SymbolicRegression", "DynamicExpressions"]; sysimage_path="SymbolicRegression.so", precompile_execution_file="precompile_file.jl")'
```

which can then be loaded with
```bash
julia -O3 --threads=auto -J SymbolicRegression.so
```

To take full advantage of the improved startup time, `Options` should be loaded with `define_helper_functions=false`.

---

This also fixes an issue where equations with complexity 0 (say you define constants to have 0 complexity) were causing bounds errors.

---

To-do:

- [x] Check if type specialization to parallelism type can be turned off.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ build
 *.csv.out*
 pysr_recorder.json
 docs/src/index.md
+*.code-workspace
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SymbolicRegression"
 uuid = "8254be44-1295-4e6a-a16d-46603ac705cb"
 authors = ["MilesCranmer <miles.cranmer@gmail.com>"]
-version = "0.14.4"
+version = "0.14.5"
 
 [deps]
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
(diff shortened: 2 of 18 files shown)


MilesCranmer:
cc @AlCap23 @ChrisRackauckas this should speed up the startup time of the search in DataDrivenDiffEq.jl quite a bit too.

@AlCap23 for faster startup, I would change the API so that `SymbolicRegression.Options` is passed a vector of functions, rather than a tuple. It no longer needs to specialize to a tuple of functions, and so compilation is a lot quicker now.

PR #165, opened 2022-12-23
[WIP] Fix Windows CI

MilesCranmer:
This eliminates a few potential data races in an attempt to fix https://github.com/JuliaLang/julia/issues/47957. I still see segfaults on Windows for earlier Julia versions, so am going to try ThreadSanitizer too. If I can't solve it on the SymbolicRegression.jl side, perhaps it is something else. Regardless, these data races should be fixed anyways (even though they don't seem to affect the search).


MilesCranmer:
Seems like this has been fixed by setting the `--heap-size-hint`

PR #175, opened 2023-02-18
Force safe closing of threads

MilesCranmer:
This fixes another potential source of bugs due to data races. In short: threads were not being killed, even if the search was stopped early. This adds a simple `wait` to each thread (which is what `rmprocs` already does for processes) at the end of `EquationSearch`.


MilesCranmer:
I think this needs to check whether `allPops[i][j]` has been started or not?

PR #183, opened 2023-03-17
Abstract number support

MilesCranmer:
This PR adds support for data type `<:Number` following the type constraints relaxation in https://github.com/SymbolicML/DynamicExpressions.jl/pull/16. This PR introduces this to the SymbolicRegression.jl side this by splitting parametric numeric types into two categories:

1. `T<:DATA_TYPE` for the data types (`X`, `y`, `weights`, the parameters inside a `tree`, and the output of a `tree(X)`). For this PR, `DATA_TYPE=Number` (the global type constraints are defined in `ProgramConstants.jl`, for easier extensions later)
2. `L<:LOSS_TYPE` for the loss types (`eval_loss(...)::L`). For this PR, `LOSS_TYPE=Real`.

This means that now the dataset object is parameterized by two types: `Dataset{T,L}`.

With this, you can pass complex data for `X` and `y`, but still optimize a real loss function. This also works with BFGS-based constant optimization!

Here's a video to show off how this works:

https://user-images.githubusercontent.com/7593028/225827839-92b7ee2d-cadf-425b-92a8-98c7bf543553.mp4

you can also just pass the parameter `loss_type` directly to `EquationSearch` if you use the `EquationSearch(X, y; loss_type=loss_type)` syntax. Also, if the type is `Complex{L}`, the loss type will be automatically inferred.

TODO:

- [x] Add integration tests for `ComplexF64`, `ComplexF32`, `ComplexF16`.
- [x] Automatically set `loss_type` when user passes complex types.
- [x] See if `Integer` works without further modification (would require `should_optimize_constants=false`).
    - (It does not. Requires more tweaks to get integer equations working... Will leave for later.)
- [x] Add documentation.
- [x] Add examples.
- [x] Get `DynamicExpressions.jl` to print complex numbers in brackets.
- [x] Do an integration test with PySR.
- [x] Do a run on the cluster to make sure distributed compute works for complex.

### DIFF
--- a/src/ConstantOptimization.jl
+++ b/src/ConstantOptimization.jl
@@ -3,15 +3,15 @@ module ConstantOptimizationModule
 using LineSearches: LineSearches
 using Optim: Optim
 import DynamicExpressions: Node, get_constants, set_constants, count_constants
-import ..CoreModule: Options, Dataset
+import ..CoreModule: Options, Dataset, DATA_TYPE, LOSS_TYPE
 import ..UtilsModule: get_birth_order
 import ..LossFunctionsModule: score_func, eval_loss
 import ..PopMemberModule: PopMember
 
 # Proxy function for optimization
 function opt_func(
-    x::Vector{T}, dataset::Dataset{T}, tree::Node{T}, options::Options
-)::T where {T<:Real}
+    x::Vector{T}, dataset::Dataset{T,L}, tree::Node{T}, options::Options
+)::L where {T<:DATA_TYPE,L<:LOSS_TYPE}
     set_constants(tree, x)
     # TODO(mcranmer): This should use score_func batching.
     loss = eval_loss(tree, dataset, options)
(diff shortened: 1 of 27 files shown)


MilesCranmer:
@AlCap23 @ChrisRackauckas would one of you be up for taking a quick look at this change? My hope is that this change would make  SymbolicRegression.jl more compatible with custom numeric datatypes.

MilesCranmer:
It might be nice to automatically set the `loss_type=L` when the user passes a type: `Complex{L}`.

src/ConstantOptimization.jl
```diff
@@ -20,16 +20,19 @@ end
...
     num_evals = 0.0
     if nconst == 0
         return (member, 0.0)
     end
     x0 = get_constants(member.tree)
-    f(x::Vector{T})::T = opt_func(x, dataset, member.tree, options)
-    if nconst == 1
+    f(x::Vector{T})::L = opt_func(x, dataset, member.tree, options)
+    if T <: Complex
+        # TODO: Make this more general. Also, do we even need Newton here at all??
```
ChrisRackauckas, line 33: I'm not sure why the other would be Newton. Is there a reason for preferring Newton in the Real case? With finite differences for the Hessian, I'd suspect it's not actually doing much better. If anything, doing a single Hessian to pre-start the BFGS could get similar robustness for less work.

ChrisRackauckas:
Looks like a generally good idea.

src/ConstantOptimization.jl
MilesCranmer: Thanks - If I remember correctly, I found Newton was cheaper for 1-parameter optimizations than the others, so seemed like a good default for simple equations. (It will be run multiple times throughout the search). But I don't remember specifics, I should really measure this again.

For complex numbers, of these 3, it seems like only BFGS is implemented:

```julia
using Optim

f(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2 + 1im)^2

x0 = zeros(ComplexF64, 2)

optimize(x -> abs2(f(x)), x0, Newton())
# MethodError: no method matching eps(::Type{ComplexF64})
optimize(x -> abs2(f(x)), x0, NelderMead())
# MethodError: no method matching isless(::ComplexF64, ::ComplexF64)

optimize(x -> abs2(f(x)), x0, BFGS())
# Works!
```

---

(Btw, another to-do I have is to allow the user to pass an arbitrary optimizer.)

ChrisRackauckas: ahh yeah for one parameter optimizations or very small then Newton can be fine. With complex, the Hessian is a bit hard to define because it's dependent on if the function is analytic or not (in the non-analytic case it needs to have a much larger system), so I don't think that will be solved any time soon. So BFGS is going to be needed with the complex case.

MilesCranmer:
Thanks Chris! 

Everything seems to be working. Merging...

PR #190, opened 2023-03-25
Multiple refactors: arbitrary data in `Dataset`, separate mutation weight conditioning, fix data races, cleaner API

MilesCranmer:
This refactors a few different internals:

1. Expands the `Dataset` struct to have an `extra::NamedTuple` field, that can be used by the user to pass basically any data to use in a custom objective. This also allows `y` to be `nothing` (only when passed as a `Dataset` object).
1. `calculate_pareto_frontier(X, y, hall_of_fame, options)` is now just `calculate_pareto_frontier(hall_of_fame)`. (The old method will still work exactly the same, but will eventually deprecated).
1. Mutation weight conditioning has been refactored into its own function. This is to enable future approaches, like with reinforcement learning, to easily change the mutation weights at runtime.
1. Prevents a few data races in population copies.


docs/src/examples.md
```diff
@@ -10,8 +10,8 @@ using DataFrames
 We'll also code up a simple function to print a single expression:
 
 ```julia
-function get_best(; X, y, hof::HallOfFame{T,L}, options) where {T,L}
-    dominating = calculate_pareto_frontier(X, y, hof, options)
+function get_best(; hof::HallOfFame{T,L}, options) where {T,L}
+    dominating = calculate_pareto_frontier(hof, options)
```
MilesCranmer, line 14: /wtd Remove `options` from this call

MilesCranmer:
Small improvements in speed across the board, despite the extra copies:


|                       | master        | refactor7     | t[master]/t[refactor7] |
|:----------------------|:-------------:|:-------------:|:----------------------:|
| search/multithreading | 18.4 ± 1.6 s  | 16.8 ± 0.76 s | 1.1                    |
| search/serial         | 52 ± 0.044 s  | 51.5 ± 0.14 s | 1.01                   |
| time_to_load          | 3.45 ± 0.14 s | 3.35 ± 0.25 s | 1.03                   |



plot_SymbolicRegression

PR #210, opened 2023-05-09
Cache complexity of expressions whenever possible

MilesCranmer:
As reported on https://github.com/MilesCranmer/PySR/discussions/325, one of the major bottlenecks is complexity calculations. It is recomputed frequently in the code because it is a cheap operation, but it seems to have added up. Also the speed up of the expression evaluation makes it more apparent.

This caches complexity of expressions, and attempts to avoid recomputing the complexity whenever possible.


MilesCranmer:
- [ ] Since this PR only implements caching, and should not change the actual algorithm, one test I should try is to make sure the deterministic output is identical before and after the PR.

MilesCranmer:
😳

MilesCranmer:
Too bad; it looks like the deterministic runs are slightly different before/after this patch. So there is some issue lurking here.

MilesCranmer:
It might be okay actually:

1. If I *always* recompute complexity, the results are the same as if I used the cached values.
2. However, if I purposefully cache an incorrect complexity and use that, the results change.

Not really sure why the output population changes before/after this PR, though.

MilesCranmer:
A better idea is to simply add
```julia
    if compute_complexity(member.tree, options) != complexity
        throw(error("Complexity is out of date."))
    end
```
to the code for checking the cache (slower, but just debugging code).

MilesCranmer:
Bug fixed. Also the bug was *partly* the cause of the massive increase in performance; it will still give a nice bump but not as significant.

MilesCranmer:
I think we don’t need a separate `assign_tree!`, can just use the behavior for `setproperty!`.

MilesCranmer:
Going to run one more test that deterministic runs are unchanged. This time I will use a custom complexity definition, and also use `weight_optimize > 0`, which should test other pathways. With that I think this is good to merge.

PR #217, opened 2023-05-28
Fix type inference problem in EquationSearch

MilesCranmer:


MilesCranmer:
Hm, not much of a performance difference and hurts readability, so I’m closing this…

PR #220, opened 2023-06-04
Dimensional constraints

MilesCranmer:
This adds what is probably the most-requested feature of all time: dimensional analysis.

https://en.wikipedia.org/wiki/Dimensional_analysis

This PR implements this by creating a ~~Unitful.jl~~ DynamicQuantities.jl ~~extension~~ which defines `violates_dimensional_constraints`. This function checks if a given equation is dimensionally consistent, given any choice of units for the free constants. (It does this by creating a "wildcard" unit that propagates through an equation until dimensionless input is needed).

By default the loss penalty for violating these constraints is 1000, but it is user configurable (`Options.dimensional_constraint_penalty`).

## Example: Newton's law

```julia
using DynamicQuantities
using SymbolicRegression

M = 1e25 .* rand(100) .* u"kg"
m = 100 .* rand(100) .* u"kg"
r = 10_000 .* rand(100) .* u"km"
G = 6.6743e-11u"m^3 * kg^−1 * s^-2"

F = @. (G * M * m / r^2)

X = ustrip.(cat(M, m, r; dims=2))'
y = ustrip.(F)

X_units = dimension.(first.([M, m, r]))  # ["kg", "kg", "m"]
y_units = dimension(first(F))  # "m kg s^-2"

function loss_fnc(prediction, target)
    # Useful loss for large dynamic range
    scatter_loss = abs(log((abs(prediction)+1e-20) / (abs(target)+1e-20)))
    sign_loss = 10 * (sign(prediction) - sign(target))^2
    return scatter_loss + sign_loss
end

options = Options(;
    binary_operators=[+, -, *, /],
    unary_operators=[cos, exp],
    maxsize=30,
    elementwise_loss=loss_fnc,
    adaptive_parsimony_scaling=1000.0,
    npopulations=Threads.nthreads() * 2,
    complexity_of_constants=2,
)

equation_search(
    X,
    y;
    niterations=1000,
    options=options,
    variable_names=["M", "m", "r"],
    units=(X=X_units, y=y_units),
)
```

https://github.com/MilesCranmer/SymbolicRegression.jl/assets/7593028/4a6d12c9-c4e9-4262-8c4f-c91b49342ae4

~~Right now the use of this is quite slow. I think this is due to the fact that Unitful.jl defines every unit as a different type, which causes the type inference to encounter a lot of instability when evaluating.~~

Fixed by switching to DynamicQuantities.jl

---

TODO:

- [x] Improve performance.
    - [x] It seems like most of the slowness was from using a `try-catch` by default to check whether an operator could handle dimension-full input. Now I first check with `hasmethod` which is significantly faster. Perhaps I will just need to advise users who use dimensional constraints to avoid passing generic operators (unless they want their operators to manipulate dimensions).
    - [x] Probably want to try this trick: https://discourse.julialang.org/t/performance-of-hasmethod-vs-try-catch-on-methoderror/99827/12
- [x] ~~`@nospecialize` some functions. Because we can encounter arbitrary equations, we can encounter arbitrary quantity types! This makes multiple dispatch explode.~~
    - ~~Is this hopeless? Maybe we need a fork of `Unitful.jl` that has a single generic `Quantity{T}`, with a *value* for the dimensions?~~
- [x] Switch to using DynamicQuantities.jl 
- [x] Refactor `score_func_batch` so that batched evaluation can use it too.
- [x] Move DynamicQuantities.jl into main package (not an extension)
- [x] Print character indicating units in constants?
- [x] ~~Scale `X` by the `upreferred` re-scalings. e.g., `u"cm"` should be scaled relative to `u"m"`.~~
    - [x] Just raising a warning instead
- [x] Require desired output units. (Need some API specification for this)
- [ ] Add dimensional analysis section to docs
- [x] Update to DynamicQuantities.jl 0.5
- [ ] Make the printout read out the original units the user passes, as right now it is unclear.


MilesCranmer:
@tfiers (or maybe @timholy?) do you know if there is a way to use Unitful.jl but still have type stability? Right now I am defining this "wildcard" type:

```julia
Base.@kwdef struct WildcardDimensionWrapper{T}
    val::Quantity{T} = one(Quantity{T})
    wildcard::Bool = false
    violates::Bool = false
end
```

which stores an evaluation of an expression, with units, at any node in the expression. However, since `Quantity` actually has three parameters instead of 1, and the two latter parameters can change at each node in the expression, Cthulhu.jl seems to light up with type instability flags. (So I think this is why the evaluation is slow.)

Any ideas?

MilesCranmer:
Made a package for this PR: https://github.com/SymbolicML/DynamicQuantities.jl. With this package, things are fairly quick now.

src/OptionsStruct.jl
```diff
@@ -120,16 +120,22 @@ function ComplexityMapping(;
...
-    CT,OPT<:Optim.Options,EL<:Union{SupervisedLoss,Function},FL<:Union{Nothing,Function},W
+    CT,
+    OP<:AbstractOperatorEnum,
+    OPT<:Optim.Options,
+    EL<:Union{SupervisedLoss,Function},
+    FL<:Union{Nothing,Function},
+    W,
 }
-    operators::AbstractOperatorEnum
+    operators::OP
```
MilesCranmer, line 130: Not sure this is needed

MilesCranmer:
Closed in favor of #228

PR #226, opened 2023-07-03
MLJ Integration

### LINKED ISSUE #225: [Feature]: MLJ integration
### Feature Request

I think it could be really cool to have an [MLJ](https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/) integration for SymbolicRegression. As an example of how it could be used, in https://discourse.julialang.org/t/ann-sirus-jl-v1-2-interpretable-machine-learning-via-rule-extraction/100932/2 I compared SymbolicRegression to [SIRUS](https://github.com/rikhuijzer/SIRUS.jl) for performing interpretable regression on a basic ML example dataset (Boston housing).

MilesCranmer:
@ericphanson I tried an initial implementation. How is this?

Potentially fixes #225 

The current downsides are: 
- no way of passing custom data to the model (possible with the `extra` parameter via `SymbolicRegression.Dataset(...; extra::NamedTuple=...)`). Maybe this just isn't suitable for MLJ though.
- No way to pass `variable_names`, or, once #220 merges, pass `units`. Again, perhaps this just is not MLJ-compatible so we don't need to worry about it.

Here's a demo:

```julia
julia> model = SRRegressor(binary_operators=[+, *, -, /], unary_operators=[cos], niterations=200, maxsize=20, nested_constraints=[cos=>[cos=>0]], complexity_of_operators=[cos=>2])
...

julia> X = rand(5, 1000) .* 15 .- 7.5;

julia> y = @. cos(X[3, :] * 2.1 - 0.2) + 0.2 * X[2, :]^2 * X[1, :];

julia> mach = machine(model, X, y)
...

julia> fit!(mach)
...

julia> report(mach)
(best_idx = 8,
 equations = DynamicExpressions.EquationModule.Node{Float64}[x1, (x1 * 3.740921188733064), ((3.7285630857630214 * x1) - -0.9321906358198151), (((x2 * x2) * x1) * 0.19990596875363167), ((((x2 * x1) + 0.03426565943145631) * 0.19989890752841208) * x2), ((((x2 * x2) * x1) + (0.01871055082023076 / x3)) * 0.19990596872137176), ((((x2 * x2) * x1) * 0.19991301441424844) + cos(x3 / -0.47083256472261414)), ((((x2 * x2) * x1) * 0.19991301441424844) + cos((x3 / -0.47083256472261414) - -0.1750279536531524))],
 equation_strings = ["x1", "(x1 * 3.740921188733064)", "((3.7285630857630214 * x1) - -0.9321906358198151)", "(((x2 * x2) * x1) * 0.19990596875363167)", "((((x2 * x1) + 0.03426565943145631) * 0.19989890752841208) * x2)", "((((x2 * x2) * x1) + (0.01871055082023076 / x3)) * 0.19990596872137176)", "((((x2 * x2) * x1) * 0.19991301441424844) + cos(x3 / -0.47083256472261414))", "((((x2 * x2) * x1) * 0.19991301441424844) + cos((x3 / -0.47083256472261414) - -0.1750279536531524))"],
 losses = [352.4189784722068, 210.99011897736824, 210.12401468032897, 0.5091640874420742, 0.5082645089530076, 0.5069986513792896, 0.024084496999990993, 0.005418666193545019],
 complexities = [1, 3, 5, 7, 9, 11, 13, 15],
 scores = [-5.864820747527641, 0.25650472223596993, 0.0020567001387135193, 3.01134140171983, 0.0008841688504815699, 0.001246827648481203, 1.523469992992648, 0.7458593298183651],)

julia> report(mach).equation_strings
8-element Vector{String}:
 "x1"
 "(x1 * 3.740921188733064)"
 "((3.7285630857630214 * x1) - -0.9321906358198151)"
 "(((x2 * x2) * x1) * 0.19990596875363167)"
 "((((x2 * x1) + 0.03426565943145631) * 0.19989890752841208) * x2)"
 "((((x2 * x2) * x1) + (0.01871055082023076 / x3)) * 0.19990596872137176)"
 "((((x2 * x2) * x1) * 0.19991301441424844) + cos(x3 / -0.47083256472261414))"
 "((((x2 * x2) * x1) * 0.19991301" ⋯ 37 bytes ⋯ "261414) - -0.1750279536531524))"

julia> report(mach).best_idx
8

julia> ypred = predict(mach, X);

julia> sum(abs2,  ypred .- y) ./ length(y)
0.005418666193545014
```

The default selection going into `.best_idx` is set by `model.selection_method` which returns an index for chosen equation.

---

@moelf is this the right approach for MLJ or would it be better to have a separate package? MLJModelInterface.jl looks pretty lightweight.


MilesCranmer:
@ablaom I'd be very interested to hear your thoughts on this implementation as well.

ablaom:
Thanks for working on an MLJ integration. Very exciting to have new kind class of model, and implemented in Julia, right?

Currently on leave but will definitely be support this. @OkonSamuel may also be able to look at this.

On a quick scan, I couldn't find any trait definitions (aka metadata) as defined, e.g. using `MMI.metadata_model` and `MMI.metadata_pkg` - see [these docs](https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#Trait-declarations). But maybe I missed them...

And it will be good idea to add `MLJTestInterface.jl` tests to your tests. There's an example for [XGBoost](https://github.com/JuliaAI/MLJXGBoostInterface.jl/blob/e3c3b5520df427d3b62edc61b3acb14549a6b362/test/runtests.jl#L205C25-L205C25) you can probably adapt.

MilesCranmer:
Fantastic, thank you. Will add the metadata.

Yes, this package is 100% Julia. The Python frontend [PySR](https://github.com/MilesCranmer/PySR) is just a lightweight scikit-learn wrapper via PyJulia, but the engine is all on the Julia side. So far the existing Julia interface in SymbolicRegression.jl is not as user-friendly, so I am looking forward to adding this MLJ integration as I think it will certainly improve that, as well as extensibility.

ericphanson:
> 2\. No way to pass `variable_names`

For this, I think a natural way in MLJ would be to allow users to pass `X` (or rather its transpose) as a table and to use the column names, since MLJ generally works with tables where rows are observations and columns are features: https://alan-turing-institute.github.io/MLJ.jl/stable/getting_started/#Two-dimensional-data

src/MLJInterface.jl
```diff
@@ -0,0 +1,163 @@
...
+) where {F,T,N<:Node{T},VN<:Vector{N},MN<:Vector{VN}}
+    return [
+        dispatch_selection(
+            selection_method, trees[i], losses[i], scores[i], complexities[i]
+        ) for i in eachindex(trees)
+    ]
+end
+
+# TODO: Should somehow automatically document kws available.
+function SRRegressor(;
```
OkonSamuel, line 54: The function keyword arguments should correspond to the struct keys.

OkonSamuel:
@MilesCranmer I have taken a first pass through the code and given my review above.
Asides from the suggestions made by @ablaom you also need to add docstring for the MLJ interface. See [here](https://alan-turing-institute.github.io/MLJ.jl/stable/adding_models_for_general_use/#Document-strings)
I'll go through again after you add MLJ Interface tests.

Moelf:
I just want to say that having integration here or a separate packages are both fine, the latter is not uncommon:
- https://github.com/FluxML/MLJFlux.jl
- https://github.com/JuliaAI/MLJXGBoostInterface.jl

src/MLJInterface.jl
MilesCranmer: The dataset is documented here: https://astroautomata.com/SymbolicRegression.jl/stable/types/#SymbolicRegression.CoreModule.DatasetModule.Dataset. The user normally does not need to worry about this though as it is for internal use, for storing the variable names, units, normalization, etc.

MilesCranmer: The dataset is useful if the user wants to define a complex custom objective; they can pass any other data their function needs to the `extra` parameter (3D tensors, graphs, etc). But this is uncommon and probably incompatible with other ML algorithms.

MilesCranmer: > This is especially important for model hyperparameter tuning

I see, thanks. So one option is for us to generate the SRRegressor struct using the keywords available to the `Options` constructor. And then, inside `fit`, the normal `Options` can be created to be passed to `equation_search`. (Some keywords in the `Options` constructor are different from the stored keys)

MilesCranmer:
Thanks. I wonder if those were created before the lightweight interface package was created? It looks like the time-to-load is negligible ( https://github.com/MilesCranmer/SymbolicRegression.jl/pull/226#issuecomment-1617346161 ) so we could just make it a hard dependency for simplicity.

ablaom:
In case this is relevant to the discussion about the hyperparmater struct (aka "model") (I didn't follow the details): The actual struct fields do not matter. The MLJ interface only interacts with the struct through *properties*,  which can be overloaded to be different from the fields. But, yes the (public) keyword constructor should match the properties, if these are different from the fields. 

For example, in MLJ's `Pipeline` model struct, we have a field `models`, invisible to the user, which is a vector of the component models, but the user constructs names for individual models (which become properties) using the kwarg constructor. I say this, because the layering mentioned above and copied below looks a bit weird, and I expect it can be avoided by having properties (public API) different from fields (implementation detail):

```julia
SRR(;options = SymbolicRegression.Options(
    unary_operators=[cos, exp],
    npopulations=20
))
```

MilesCranmer:
Okay I think I found a nice way around this. The way it is implemented is far from elegant, but the user point of view is nice.

It uses some macro tricks to store all the parameters and defaults in `Options`, and use those to generate the fields and defaults of `SRRegressor`. So basically the user no longer needs to declare `Options()` if they are working with MLJ.

I thought about just using `options` as a keyword but I felt it made things a bit awkward, and maybe it would cause issues for hyperparameter tuning.

<details><summary>
For better or for worse, now you get all hyperparameters directly in the SRRegressor:
</summary>

```julia
julia> model = SRRegressor(binary_operators=[+, *, -, /], unary_operators=[cos], maxsize=25, nested_constraints=[cos=>[cos=>0]], niterations=100)
SRRegressor(
  binary_operators = Function[+, *, -, /], 
  unary_operators = [cos], 
  constraints = nothing, 
  elementwise_loss = nothing, 
  loss_function = nothing, 
  tournament_selection_n = 12, 
  tournament_selection_p = 0.86f0, 
  topn = 12, 
  complexity_of_operators = nothing, 
  complexity_of_constants = nothing, 
  complexity_of_variables = nothing, 
  parsimony = 0.0032f0, 
  alpha = 0.1f0, 
  maxsize = 25, 
  maxdepth = nothing, 
  fast_cycle = false, 
  turbo = false, 
  migration = true, 
  hof_migration = true, 
  should_simplify = nothing, 
  should_optimize_constants = true, 
  output_file = nothing, 
  npopulations = 15, 
  perturbation_factor = 0.076f0, 
  annealing = false, 
  batching = false, 
  batch_size = 50, 
  mutation_weights = MutationWeights(0.048, 0.47, 0.79, 5.1, 1.7, 0.002, 0.00023, 0.21, 0.0), 
  crossover_probability = 0.066f0, 
  warmup_maxsize_by = 0.0f0, 
  use_frequency = true, 
  use_frequency_in_tournament = true, 
  adaptive_parsimony_scaling = 20.0, 
  npop = 33, 
  ncycles_per_iteration = 550, 
  fraction_replaced = 0.00036f0, 
  fraction_replaced_hof = 0.035f0, 
  verbosity = 1000000000, 
  save_to_file = true, 
  probability_negate_constant = 0.01f0, 
  seed = nothing, 
  bin_constraints = nothing, 
  una_constraints = nothing, 
  progress = true, 
  terminal_width = nothing, 
  optimizer_algorithm = "BFGS", 
  optimizer_nrestarts = 2, 
  optimizer_probability = 0.14f0, 
  optimizer_iterations = nothing, 
  optimizer_options = nothing, 
  recorder = nothing, 
  recorder_file = "pysr_recorder.json", 
  early_stop_condition = nothing, 
  timeout_in_seconds = nothing, 
  max_evals = nothing, 
  skip_mutation_failures = true, 
  enable_autodiff = false, 
  nested_constraints = [cos => [cos => 0]], 
  deterministic = false, 
  define_helper_functions = true, 
  niterations = 100, 
  parallelism = :multithreading, 
  numprocs = nothing, 
  procs = nothing, 
  addprocs_function = nothing, 
  runtests = true, 
  loss_type = Nothing, 
  selection_method = SymbolicRegression.MLJInterfaceModule.choose_best)
```

</details>

The downside is it has to re-create the `Options()` struct whenever you call, e.g., `fit` or `predict`. I think it should be fine though. We could even store `options` in the `fitresult` to avoid invalidating the returned expressions (e.g., if the user tries to change the operators used; you would still want `predict()` to work with the old result).

There are a couple of nested hyperparameters, for example `MutationWeights` is a struct storing some probabilities. I'm not sure if that would pose an issue for hyperparam tuning or not?

src/MLJInterface.jl
```diff
@@ -0,0 +1,204 @@
...
+MMI.package_uuid(::Type{<:AbstractSRRegressor}) = "8254be44-1295-4e6a-a16d-46603ac705cb"
+function MMI.package_url(::Type{<:AbstractSRRegressor})
+    return "https://github.com/MilesCranmer/SymbolicRegression.jl"
+end
+MMI.package_license(::Type{<:AbstractSRRegressor}) = "Apache-2.0"
+MMI.is_pure_julia(::Type{<:AbstractSRRegressor}) = true
+MMI.is_wrapper(::Type{<:AbstractSRRegressor}) = false
+
+MMI.input_scitype(::Type{SRRegressor}) = MMI.Table(MMI.Continuous)
+MMI.target_scitype(::Type{SRRegressor}) = AbstractVecOrMat{MMI.Continuous}
```
MilesCranmer, line 162: Is this the appropriate trait to declare to indicate that `SRRegressor` can handle multi-output targets?

src/MLJInterface.jl
```diff
@@ -0,0 +1,203 @@
...
+    options = get_options(m)
+    X_t = transpose(MMI.matrix(X))
+    # TODO: Is this needed? Would MLJ ever pass in a matrix for y?
+    y_t = ndims(y) == 1 ? transpose(MMI.matrix(transpose(y))) : transpose(MMI.matrix(y))
+    search_state = equation_search(
+        X_t,
+        y_t;
+        niterations=m.niterations,
+        weights=w,
+        variable_names=nothing,
```
MilesCranmer, line 109: I could try using @ericphanson's idea to get `variable_names` from an input table. Is there a function available for getting this? I looked at using `keys` but apparently `keys(::AbstractArray)` returns the indices by default.

src/MLJInterface.jl
```diff
@@ -0,0 +1,203 @@
...
+MMI.package_uuid(::Type{<:AbstractSRRegressor}) = "8254be44-1295-4e6a-a16d-46603ac705cb"
+function MMI.package_url(::Type{<:AbstractSRRegressor})
+    return "https://github.com/MilesCranmer/SymbolicRegression.jl"
+end
+MMI.package_license(::Type{<:AbstractSRRegressor}) = "Apache-2.0"
+MMI.is_pure_julia(::Type{<:AbstractSRRegressor}) = true
+MMI.is_wrapper(::Type{<:AbstractSRRegressor}) = false
+
+MMI.input_scitype(::Type{SRRegressor}) = MMI.Table(MMI.Continuous)
+MMI.target_scitype(::Type{SRRegressor}) = AbstractVecOrMat{<:MMI.Continuous}
```
MilesCranmer, line 161: Is this the appropriate trait to declare to indicate that SRRegressor can handle multi-output targets?

It still seems like I get a warning if I try to fit on multi-dimensional outputs

src/MLJInterface.jl
```diff
@@ -0,0 +1,211 @@
...
+    return _update(m, verbosity, (; state=nothing), nothing, X, y, w)
+end
+function MMI.update(
+    m::AbstractSRRegressor, verbosity, old_fitresult, old_cache, X, y, w=nothing
+)
+    return _update(m, verbosity, old_fitresult, old_cache, X, y, w)
+end
+function _update(m::AbstractSRRegressor, verbosity, old_fitresult, old_cache, X, y, w)
+    options = get(old_fitresult, :options, get_options(m))
+    X_t = transpose(MMI.matrix(X))
```
MilesCranmer, line 109: Note to self: should ensure that this makes a copy rather than a view. Would want to avoid accessing feature slices of data in row-major format as it would be slower.

src/MLJInterface.jl
OkonSamuel: In the case of an abstract array its better to assign default keys. something like
`[Symbol("x$i") for i in 1:5]`
This is done in several other interface packages.

src/MLJInterface.jl
OkonSamuel: It's better to split your model into two.
One that handles single target. The other handles multitarget.
e.r `SRRegressor` for single target and `MultiSRRegressor` for multi-target.

OkonSamuel: Also rather than defining these traits by hand. You can use the helper function
`metadata_model` to define it and its much cleaner see https://github.com/JuliaAI/MLJMultivariateStatsInterface.jl/blob/dev/src/models/linear_models.jl#L148

src/MLJInterface.jl
MilesCranmer: Right, this is what already happens inside `equation_search`. But I guess I am wondering what the general strategy is for extracting keys if they **do** exist in the input data. Like `NamedTuple` or `Table` types for example.

src/MLJInterface.jl
MilesCranmer: Good idea, I will switch to that.

src/MLJInterface.jl
```diff
@@ -0,0 +1,214 @@
...
+MMI.package_uuid(::Type{<:AbstractSRRegressor}) = "8254be44-1295-4e6a-a16d-46603ac705cb"
+function MMI.package_url(::Type{<:AbstractSRRegressor})
+    return "https://github.com/MilesCranmer/SymbolicRegression.jl"
+end
+MMI.package_license(::Type{<:AbstractSRRegressor}) = "Apache-2.0"
+MMI.is_pure_julia(::Type{<:AbstractSRRegressor}) = true
+MMI.is_wrapper(::Type{<:AbstractSRRegressor}) = false
+
+MMI.input_scitype(::Type{SRRegressor}) = MMI.Table(MMI.Continuous)
+function MMI.target_scitype(::Type{SRRegressor})
```
OkonSamuel, line 170: Split the model into two. see above comment

MilesCranmer: Thanks; done!

src/MLJInterface.jl
MilesCranmer: Done! Added the MultiSRRegressor.

I tried out the `metadata_model` generator but it seems perhaps too abstracted away. The simple methods-based version seems a bit clearer to me. Is it okay from an implementation point of view if I stick with the methods-based version? I understand it is a bit more tedious but it is a bit more readable for me for whatever reason...

src/MLJInterface.jl
```diff
@@ -0,0 +1,214 @@
...
+    options = get(old_fitresult, :options, get_options(m))
+    X_t = transpose(MMI.matrix(X))
+    # TODO: Is this needed? Would MLJ ever pass in a matrix for y?
+    y_t = ndims(y) == 1 ? transpose(MMI.matrix(transpose(y))) : transpose(MMI.matrix(y))
+    search_state = equation_search(
+        X_t,
+        y_t;
+        niterations=m.niterations,
+        weights=w,
+        variable_names=nothing,
```
OkonSamuel, line 113:
```julia
sch = MMI.schema(X)
variable_names = getcolnames(sch, Xmatrix)
getcolnames(::Nothing, Xmatrix)= [[Symbol("x$i") for i in 1:size(Xmatrix, 2)]
getcolnames(sch, Xmatrix) = sch.names 
```

src/MLJInterface.jl
OkonSamuel: Yes, at the moment we haven't yet implemented a standard way of doing. I will open a PR in the future to address this.
But for now you can use the code below to do this,

src/MLJInterface.jl
MilesCranmer: Good idea; let me add that

src/MLJInterface.jl
```diff
@@ -0,0 +1,211 @@
...
+        isa(s, Symbol) && return s
+    end
+    return nothing
+end
+
+"""Get an equivalent `Options()` object for a particular regressor."""
+function get_options(::AbstractSRRegressor) end
+
+eval(modelexpr(:SRRegressor))
+eval(modelexpr(:MultiSRRegressor))
```
OkonSamuel, line 70:
```suggestion
eval(modelexpr(:MultitargetSRRegressor))
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,211 @@
...
+        runtests=m.runtests,
+        saved_state=old_fitresult.state,
+        return_state=true,
+        loss_type=m.loss_type,
+    )
+    fitresult = (; state=search_state, options=options)
+    return (fitresult, nothing, full_report(m, fitresult))
+end
+format_input_for(::SRRegressor, y) = transpose(MMI.matrix(transpose(y)))
+format_input_for(::MultiSRRegressor, y) = transpose(MMI.matrix(y))
```
OkonSamuel, line 128:
```suggestion
format_input_for(::MultitargetSRRegressor, y) = transpose(MMI.matrix(y))
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,211 @@
...
+        addprocs_function=m.addprocs_function,
+        runtests=m.runtests,
+        saved_state=old_fitresult.state,
+        return_state=true,
+        loss_type=m.loss_type,
+    )
+    fitresult = (; state=search_state, options=options)
+    return (fitresult, nothing, full_report(m, fitresult))
+end
+format_input_for(::SRRegressor, y) = transpose(MMI.matrix(transpose(y)))
```
OkonSamuel, line 127:
```suggestion
format_input_for(::SRRegressor, y) = y
```
I taught for this case `y` was a vector. No?

MilesCranmer: It is, yeah. I think I did this because `y` was being passed in as a weird variable type and `MMI.matrix` was helping it up (but there is no equivalent `MMI.vector`). But let me double check if it's even needed.

src/MLJInterface.jl
MilesCranmer: Hm, it seems like `MMI.matrix` converts it to a regular `Matrix` which means that `MMI.schema` no longer applies. Maybe `MMI.table` is the right thing to use instead?

MilesCranmer: Okay I think it is fixed with `MMI.table` and then your `getcolnames`. Thanks!

src/MLJInterface.jl
```diff
@@ -0,0 +1,215 @@
...
+MMI.supports_weights(::Type{<:AbstractSRRegressor}) = true
+MMI.reports_feature_importances(::Type{<:AbstractSRRegressor}) = false
+
+MMI.target_scitype(::Type{SRRegressor}) = AbstractVector{<:MMI.Continuous}
+MMI.load_path(::Type{SRRegressor}) = "SymbolicRegression.MLJInterfaceModule.SRRegressor"
+MMI.human_name(::Type{SRRegressor}) = "Symbolic Regression via Evolutionary Search"
+
+MMI.target_scitype(::Type{MultiSRRegressor}) = MMI.Table(MMI.Continuous)
+MMI.load_path(::Type{MultiSRRegressor}) = "SymbolicRegression.MLJInterfaceModule.MultiSRRegressor"
+MMI.human_name(::Type{MultiSRRegressor}) = "Multi-Target Symbolic Regression via Evolutionary Search"
```
MilesCranmer, line 181:
```suggestion
MMI.target_scitype(::Type{MultitargetSRRegressor}) = MMI.Table(MMI.Continuous)
MMI.load_path(::Type{MultitargetSRRegressor}) = "SymbolicRegression.MLJInterfaceModule.MultitargetSRRegressor"
MMI.human_name(::Type{MultitargetSRRegressor}) = "Multi-Target Symbolic Regression via Evolutionary Search"
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,215 @@
...
+end
+function MMI.predict(m::SRRegressor, fitresult, Xnew)
+    params = MMI.fitted_params(m, fitresult)
+    Xnew_t = transpose(MMI.matrix(Xnew))
+    eq = params.equations[params.best_idx]
+    out, flag = eval_tree_array(eq, Xnew_t, fitresult.options)
+    !flag && error("Detected a NaN in evaluating expression.")
+    return out
+end
+function MMI.predict(m::MultiSRRegressor, fitresult, Xnew)
```
MilesCranmer, line 149:
```suggestion
function MMI.predict(m::MultitargetSRRegressor, fitresult, Xnew)
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,215 @@
...
+
+MMI.target_scitype(::Type{MultiSRRegressor}) = MMI.Table(MMI.Continuous)
+MMI.load_path(::Type{MultiSRRegressor}) = "SymbolicRegression.MLJInterfaceModule.MultiSRRegressor"
+MMI.human_name(::Type{MultiSRRegressor}) = "Multi-Target Symbolic Regression via Evolutionary Search"
+#! format: on
+
+function get_equation_strings_for(::SRRegressor, trees, options)
+    return (t -> string_tree(t, options)).(trees)
+end
+function get_equation_strings_for(::MultiSRRegressor, trees, options)
```
MilesCranmer, line 187:
```suggestion
function get_equation_strings_for(::MultitargetSRRegressor, trees, options)
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,215 @@
...
+        (losses[i] <= threshold) ? scores[i] : typemin(L) for i in eachindex(losses)
+    ])
+end
+
+function dispatch_selection_for(m::SRRegressor, trees, losses, scores, complexities)
+    return m.selection_method(;
+        trees=trees, losses=losses, scores=scores, complexities=complexities
+    )::Integer
+end
+function dispatch_selection_for(m::MultiSRRegressor, trees, losses, scores, complexities)
```
MilesCranmer, line 207:
```suggestion
function dispatch_selection_for(m::MultitargetSRRegressor, trees, losses, scores, complexities)
```

src/SymbolicRegression.jl
```diff
@@ -8,6 +8,8 @@ export Population,
     Dataset,
     MutationWeights,
     Node,
+    SRRegressor,
+    MultiSRRegressor,
```
MilesCranmer, line 12:
```suggestion
    MultitargetSRRegressor,
```

src/SymbolicRegression.jl
```diff
@@ -1013,6 +1031,9 @@ function _equation_search(
     end
 end
 
+include("MLJInterface.jl")
+import .MLJInterfaceModule: SRRegressor, MultiSRRegressor
```
MilesCranmer, line 1035:
```suggestion
import .MLJInterfaceModule: SRRegressor, MultitargetSRRegressor
```

test/test_mlj.jl
```diff
@@ -0,0 +1,18 @@
+import SymbolicRegression: SRRegressor, MultiSRRegressor
```
MilesCranmer, line 1:
```suggestion
import SymbolicRegression: SRRegressor, MultitargetSRRegressor
```

test/test_mlj.jl
```diff
@@ -0,0 +1,18 @@
...
+    failures, summary = MTI.test(
+        [SRRegressor], MTI.make_regression()...; mod=@__MODULE__, verbosity=0, throw=true
+    )
+    @test isempty(failures)
+
+    X = randn(100, 3)
+    Y = @. cos(X^2) * 3.2 - 0.5
+    (X, Y) = MTI.table.((X, Y))
+    failures, summary = MTI.test(
+        [MultiSRRegressor], X, Y; mod=@__MODULE__, verbosity=0, throw=true
```
MilesCranmer, line 15:
```suggestion
        [MultitargetSRRegressor], X, Y; mod=@__MODULE__, verbosity=0, throw=true
```

src/MLJInterface.jl
OkonSamuel, line 149:
```suggestion
function MMI.predict(m::MultitargetSRRegressor, fitresult, Xnew)
```
So its same as in other interfaces

src/MLJInterface.jl
```diff
@@ -0,0 +1,215 @@
...
+
+MMI.input_scitype(::Type{<:AbstractSRRegressor}) = MMI.Table(MMI.Continuous)
+MMI.supports_weights(::Type{<:AbstractSRRegressor}) = true
+MMI.reports_feature_importances(::Type{<:AbstractSRRegressor}) = false
+
+MMI.target_scitype(::Type{SRRegressor}) = AbstractVector{<:MMI.Continuous}
+MMI.load_path(::Type{SRRegressor}) = "SymbolicRegression.MLJInterfaceModule.SRRegressor"
+MMI.human_name(::Type{SRRegressor}) = "Symbolic Regression via Evolutionary Search"
+
+MMI.target_scitype(::Type{MultiSRRegressor}) = MMI.Table(MMI.Continuous)
```
OkonSamuel, line 179:
```suggestion
MMI.target_scitype(::Type{MultitargetSRRegressor}) = MMI.Table(MMI.Continuous)
```

src/MLJInterface.jl
OkonSamuel:
```suggestion
MMI.load_path(::Type{MultitargetSRRegressor}) = "SymbolicRegression.MLJInterfaceModule.MultiSRRegressor"
MMI.human_name(::Type{MultitargetSRRegressor}) = "Multi-Target Symbolic Regression via Evolutionary Search"
```

src/MLJInterface.jl
OkonSamuel:
```suggestion
function dispatch_selection_for(m::MultitargetSRRegressor, trees, losses, scores, complexities)
```

src/MLJInterface.jl
OkonSamuel:
```suggestion
function get_equation_strings_for(::MultitargetSRRegressor, trees, options)
```

src/MLJInterface.jl
MilesCranmer: Thanks, added

src/MLJInterface.jl
```diff
@@ -0,0 +1,215 @@
...
+end
+function MMI.update(
+    m::AbstractSRRegressor, verbosity, old_fitresult, old_cache, X, y, w=nothing
+)
+    return _update(m, verbosity, old_fitresult, old_cache, X, y, w)
+end
+function _update(m::AbstractSRRegressor, verbosity, old_fitresult, old_cache, X, y, w)
+    options = get(old_fitresult, :options, get_options(m))
+    mX = MMI.matrix(X)
+    variable_names = getcolnames(MMI.table(X))
```
OkonSamuel, line 106: `MMI.table` will error if X is already a table.

MilesCranmer: I could have `getcolnames` call `MMI.table` if it is not already a table. What is the right type to dispatch on, for something being a table?

src/MLJInterface.jl
```diff
@@ -0,0 +1,217 @@
...
+    X_t = copy(transpose(mX))
```
OkonSamuel, line 107:
```suggestion
    X_t = permutedims(mX)
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,217 @@
...
+        loss_type=m.loss_type,
+    )
+    fitresult = (; state=search_state, options=options)
+    return (fitresult, nothing, full_report(m, fitresult))
+end
+getcolnames(X) = getcolnames(MMI.schema(X), X)
+getcolnames(::Nothing, X) = [map(i -> "x$(i)", axes(X, 2))...]
+getcolnames(sch, X) = [string.(sch.names)...]
+format_input_for(::SRRegressor, y) = y
+format_input_for(::MultitargetSRRegressor, y) = copy(transpose(MMI.matrix(y)))
```
OkonSamuel, line 132:
```suggestion
format_input_for(::MultitargetSRRegressor, y) = MMI.matrix(y, transpose=true)
```
This makes a copy

src/MLJInterface.jl
MilesCranmer: This seems to work, though it's a bit brute force:
```julia
function getcolnames(X)
    try
        return getcolnames(MMI.schema(X), X)
    catch e
        isa(e, ArgumentError) || rethrow(e)
    end
    try
        return getcolnames(MMI.schema(MMI.table(X)), X)
    catch e
        isa(e, MethodError) || isa(e, ArgumentError) || rethrow(e)
    end
    return getcolnames(nothing, X)
end
getcolnames(::Nothing, X) = [map(i -> "x$(i)", axes(X, 2))...]
getcolnames(sch, X) = [string.(sch.names)...]
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,231 @@
...
+        isa(e, ArgumentError) || rethrow(e)
+    end
+    try
+        return getcolnames(MMI.schema(MMI.table(X)), X)
+    catch e
+        isa(e, MethodError) || isa(e, ArgumentError) || rethrow(e)
+    end
+    return getcolnames(nothing, X)
+end
+getcolnames(::Nothing, X) = [map(i -> "x$(i)", axes(X, 2))...]
```
OkonSamuel, line 142: Also, the thing is that not all tables have a schema defined on the fly.

MilesCranmer: I see. So what do you think the best solution is? Will this work, because of the `try-catch`'s used above?

src/MLJInterface.jl
```diff
@@ -0,0 +1,231 @@
...
+    params = MMI.fitted_params(m, fitresult)
+    Xnew_t = transpose(MMI.matrix(Xnew))
+    eq = params.equations[params.best_idx]
+    out, flag = eval_tree_array(eq, Xnew_t, fitresult.options)
+    !flag && error("Detected a NaN in evaluating expression.")
+    return out
+end
+function MMI.predict(m::MultitargetSRRegressor, fitresult, Xnew)
+    params = MMI.fitted_params(m, fitresult)
+    Xnew_t = transpose(MMI.matrix(Xnew))
```
OkonSamuel, line 165:
```suggestion
    Xnew_t = MMI.matrix(Xnew, transpose=true)
```

I believe you intended to change this also

MilesCranmer: Yes thanks

src/MLJInterface.jl
OkonSamuel:
``` 
function get_matrix_and_colnames(X)
    sch = MMI.istable(X) ? MMI.schema(X) : nothing
    Xm_t = MMI.matrix(X, transpose=true)
    colnames = sch === nothing ? [map(i -> "x$(i)", axes(Xm_t, 1))...] : [string.(sch.names)...]
    return  Xm_t, colnames
end
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,231 @@
...
+    report = full_report(m, fitresult)
+    return (;
+        best_idx=report.best_idx,
+        equations=report.equations,
+        equation_strings=report.equation_strings,
+    )
+end
+function MMI.predict(m::SRRegressor, fitresult, Xnew)
+    params = MMI.fitted_params(m, fitresult)
+    Xnew_t = transpose(MMI.matrix(Xnew))
```
OkonSamuel, line 157:
```suggestion
    Xnew_t = MMI.matrix(Xnew, copy=true)
```

src/MLJInterface.jl
MilesCranmer: Nice, looks good!

MilesCranmer: Implemented.

src/MLJInterface.jl
```diff
@@ -0,0 +1,246 @@
...
+)
+MMI.metadata_model(
+    MultitargetSRRegressor;
+    input_scitype=Union{MMI.Table(MMI.Continuous),AbstractMatrix{<:MMI.Continuous}},
+    target_scitype=Union{MMI.Table(MMI.Continuous),AbstractMatrix{<:MMI.Continuous}},
+    supports_weights=true,
+    reports_feature_importances=false,
+    load_path="SymbolicRegression.MLJInterfaceModule.MultitargetSRRegressor",
+    descr="Multi-Target Symbolic Regression via Evolutionary Search",
+)
```
MilesCranmer, line 209: FYI @OkonSamuel I switched back to these instead 👍

MilesCranmer:
Okay I think I got rid of all the edge cases. I also added a few tests. Let me know if you have any other comments. And thanks a lot of your help thus far, it is much appreciated!! I am excited to add this to the registry.

MilesCranmer:
<details><summary>Docstring generation working too:</summary>

```julia
help?> SRRegressor
search: SRRegressor MultitargetSRRegressor

  SRRegressor


  A model type for constructing a sr regressor, based on SymbolicRegression.jl (https://github.com/MilesCranmer/SymbolicRegression.jl), and
  implementing the MLJ model interface.

  From MLJ, the type can be imported using

  SRRegressor = @load SRRegressor pkg=SymbolicRegression


  Do model = SRRegressor() to construct an instance with default hyper-parameters. Provide keyword arguments to override hyper-parameter
  defaults, as in SRRegressor(binary_operators=...).

  Arguments
  ≡≡≡≡≡≡≡≡≡≡≡

    •  binary_operators: Vector of binary operators (functions) to use. Each operator should be defined for two input scalars, and one
       output scalar. All operators need to be defined over the entire real line (excluding infinity - these are stopped before they
       are input), or return NaN where not defined. For speed, define it so it takes two reals of the same type as input, and outputs
       the same type. For the SymbolicUtils simplification backend, you will need to define a generic method of the operator so it
       takes arbitrary types.

    •  unary_operators: Same, but for unary operators (one input scalar, gives an output scalar).

    •  constraints: Array of pairs specifying size constraints for each operator. The constraints for a binary operator should be a
       2-tuple (e.g., (-1, -1)) and the constraints for a unary operator should be an Int. A size constraint is a limit to the size of
       the subtree in each argument of an operator. e.g., [(^)=>(-1, 3)] means that the ^ operator can have arbitrary size (-1) in its
       left argument, but a maximum size of 3 in its right argument. Default is no constraints.

    •  batching: Whether to evolve based on small mini-batches of data, rather than the entire dataset.

    •  batch_size: What batch size to use if using batching.

    •  elementwise_loss: What elementwise loss function to use. Can be one of the following losses, or any other loss of type
       SupervisedLoss. You can also pass a function that takes a scalar target (left argument), and scalar predicted (right argument),
       and returns a scalar. This will be averaged over the predicted data. If weights are supplied, your function should take a third
       argument for the weight scalar. Included losses: Regression: - LPDistLoss{P}(), - L1DistLoss(), - L2DistLoss() (mean square), -
       LogitDistLoss(), - HuberLoss(d), - L1EpsilonInsLoss(ϵ), - L2EpsilonInsLoss(ϵ), - PeriodicLoss(c), - QuantileLoss(τ),
       Classification: - ZeroOneLoss(), - PerceptronLoss(), - L1HingeLoss(), - SmoothedL1HingeLoss(γ), - ModifiedHuberLoss(), -
       L2MarginLoss(), - ExpLoss(), - SigmoidLoss(), - DWDMarginLoss(q).

    •  loss_function: Alternatively, you may redefine the loss used as any function of tree::Node{T}, dataset::Dataset{T}, and
       options::Options, so long as you output a non-negative scalar of type T. This is useful if you want to use a loss that takes
       into account derivatives, or correlations across the dataset. This also means you could use a custom evaluation for a
       particular expression. Take a look at _eval_loss in the file src/LossFunctions.jl for an example.

    •  npopulations: How many populations of equations to use. By default this is set equal to the number of cores

    •  npop: How many equations in each population.

    •  ncycles_per_iteration: How many generations to consider per iteration.

    •  tournament_selection_n: Number of expressions considered in each tournament.

    •  tournament_selection_p: The fittest expression in a tournament is to be selected with probability p, the next fittest with
       probability p*(1-p), and so forth.

    •  topn: Number of equations to return to the host process, and to consider for the hall of fame.

    •  complexity_of_operators: What complexity should be assigned to each operator, and the occurrence of a constant or variable. By
       default, this is 1 for all operators. Can be a real number as well, in which case the complexity of an expression will be
       rounded to the nearest integer. Input this in the form of, e.g., [(^) => 3, sin => 2].

    •  complexity_of_constants: What complexity should be assigned to use of a constant. By default, this is 1.

    •  complexity_of_variables: What complexity should be assigned to each variable. By default, this is 1.

    •  alpha: The probability of accepting an equation mutation during regularized evolution is given by exp(-delta_loss/(alpha * T)),
       where T goes from 1 to 0. Thus, alpha=infinite is the same as no annealing.

    •  maxsize: Maximum size of equations during the search.

    •  maxdepth: Maximum depth of equations during the search, by default this is set equal to the maxsize.

    •  parsimony: A multiplicative factor for how much complexity is punished.

    •  use_frequency: Whether to use a parsimony that adapts to the relative proportion of equations at each complexity; this will
       ensure that there are a balanced number of equations considered for every complexity.

    •  use_frequency_in_tournament: Whether to use the adaptive parsimony described above inside the score, rather than just at the
       mutation accept/reject stage.

    •  adaptive_parsimony_scaling: How much to scale the adaptive parsimony term in the loss. Increase this if the search is spending
       too much time optimizing the most complex equations.

    •  fast_cycle: Whether to thread over subsamples of equations during regularized evolution. Slightly improves performance, but is
       a different algorithm.

    •  turbo: Whether to use LoopVectorization.@turbo to evaluate expressions. This can be significantly faster, but is only
       compatible with certain operators. Experimental!

    •  migration: Whether to migrate equations between processes.

    •  hof_migration: Whether to migrate equations from the hall of fame to processes.

    •  fraction_replaced: What fraction of each population to replace with migrated equations at the end of each cycle.

    •  fraction_replaced_hof: What fraction to replace with hall of fame equations at the end of each cycle.

    •  should_simplify: Whether to simplify equations. If you pass a custom objective, this will be set to false.

    •  should_optimize_constants: Whether to use an optimization algorithm to periodically optimize constants in equations.

    •  optimizer_nrestarts: How many different random starting positions to consider for optimization of constants.

    •  optimizer_algorithm: Select algorithm to use for optimizing constants. Default is "BFGS", but "NelderMead" is also supported.

    •  optimizer_options: General options for the constant optimization. For details we refer to the documentation on Optim.Options
       from the Optim.jl package. Options can be provided here as NamedTuple, e.g. (iterations=16,), as a Dict, e.g. Dict(:x_tol =>
       1.0e-32,), or as an Optim.Options instance.

    •  output_file: What file to store equations to, as a backup.

    •  perturbation_factor: When mutating a constant, either multiply or divide by (1+perturbation_factor)^(rand()+1).

    •  probability_negate_constant: Probability of negating a constant in the equation when mutating it.

    •  mutation_weights: Relative probabilities of the mutations. The struct MutationWeights should be passed to these options. See
       its documentation on MutationWeights for the different weights.

    •  crossover_probability: Probability of performing crossover.

    •  annealing: Whether to use simulated annealing.

    •  warmup_maxsize_by: Whether to slowly increase the max size from 5 up to maxsize. If nonzero, specifies the fraction through the
       search at which the maxsize should be reached.

    •  verbosity: Whether to print debugging statements or not.

    •  save_to_file: Whether to save equations to a file during the search.

    •  bin_constraints: See constraints. This is the same, but specified for binary operators only (for example, if you have an
       operator that is both a binary and unary operator).

    •  una_constraints: Likewise, for unary operators.

    •  seed: What random seed to use. nothing uses no seed.

    •  progress: Whether to use a progress bar output (verbosity will have no effect).

    •  early_stop_condition: Float - whether to stop early if the mean loss gets below this value. Function - a function taking (loss,
       complexity) as arguments and returning true or false.

    •  timeout_in_seconds: Float64 - the time in seconds after which to exit (as an alternative to the number of iterations).

    •  max_evals: Int (or Nothing) - the maximum number of evaluations of expressions to perform.

    •  skip_mutation_failures: Whether to simply skip over mutations that fail or are rejected, rather than to replace the mutated
       expression with the original expression and proceed normally.

    •  enable_autodiff: Whether to enable automatic differentiation functionality. This is turned off by default. If turned on, this
       will be turned off if one of the operators does not have well-defined gradients.

    •  nested_constraints: Specifies how many times a combination of operators can be nested. For example, [sin => [cos => 0], cos =>
       [cos => 2]] specifies that cos may never appear within a sin, but sin can be nested with itself an unlimited number of times.
       The second term specifies that cos can be nested up to 2 times within a cos, so that cos(cos(cos(x))) is allowed (as well as
       any combination of + or - within it), but cos(cos(cos(cos(x)))) is not allowed. When an operator is not specified, it is
       assumed that it can be nested an unlimited number of times. This requires that there is no operator which is used both in the
       unary operators and the binary operators (e.g., - could be both subtract, and negation). For binary operators, both arguments
       are treated the same way, and the max of each argument is constrained.

    •  deterministic: Use a global counter for the birth time, rather than calls to time(). This gives perfect resolution, and is
       therefore deterministic. However, it is not thread safe, and must be used in serial mode.

    •  define_helper_functions: Whether to define helper functions for constructing and evaluating trees.

    •  niterations::Int=10: The number of iterations to perform the search. More iterations will improve the results.

    •  parallelism=:multithreading: What parallelism mode to use. The options are :multithreading, :multiprocessing, and :serial. By
       default, multithreading will be used. Multithreading uses less memory, but multiprocessing can handle multi-node compute. If
       using :multithreading mode, the number of threads available to julia are used. If using :multiprocessing, numprocs processes
       will be created dynamically if procs is unset. If you have already allocated processes, pass them to the procs argument and
       they will be used. You may also pass a string instead of a symbol, like "multithreading".

    •  numprocs::Union{Int, Nothing}=nothing: The number of processes to use, if you want equation_search to set this up
       automatically. By default this will be 4, but can be any number (you should pick a number <= the number of cores available).

    •  procs::Union{Vector{Int}, Nothing}=nothing: If you have set up a distributed run manually with procs = addprocs() and
       @everywhere, pass the procs to this keyword argument.

    •  addprocs_function::Union{Function, Nothing}=nothing: If using multiprocessing (parallelism=:multithreading), and are not
       passing procs manually, then they will be allocated dynamically using addprocs. However, you may also pass a custom function to
       use instead of addprocs. This function should take a single positional argument, which is the number of processes to use, as
       well as the lazy keyword argument. For example, if set up on a slurm cluster, you could pass addprocs_function =
       addprocs_slurm, which will set up slurm processes.

    •  runtests::Bool=true: Whether to run (quick) tests before starting the search, to see if there will be any problems during the
       equation search related to the host environment.

    •  loss_type::Type=Nothing: If you would like to use a different type for the loss than for the data you passed, specify the type
       here. Note that if you pass complex data ::Complex{L}, then the loss type will automatically be set to L.
```
</details>

src/MLJInterface.jl
```diff
@@ -0,0 +1,309 @@
...
+    descr="Symbolic Regression via Evolutionary Search",
+)
+MMI.metadata_model(
+    MultitargetSRRegressor;
+    input_scitype=Union{MMI.Table(MMI.Continuous),AbstractMatrix{<:MMI.Continuous}},
+    target_scitype=Union{MMI.Table(MMI.Continuous),AbstractMatrix{<:MMI.Continuous}},
+    supports_weights=true,
+    reports_feature_importances=false,
+    load_path="SymbolicRegression.MLJInterfaceModule.MultitargetSRRegressor",
+    descr="Multi-Target Symbolic Regression via Evolutionary Search",
```
OkonSamuel, line 241:
```suggestion
    human_name="Multi-Target Symbolic Regression via Evolutionary Search",
```

OkonSamuel: `descr` is going to be deprecated

src/MLJInterface.jl
```diff
@@ -0,0 +1,309 @@
...
+)
+
+MMI.metadata_model(
+    SRRegressor;
+    input_scitype=Union{MMI.Table(MMI.Continuous),AbstractMatrix{<:MMI.Continuous}},
+    target_scitype=AbstractVector{<:MMI.Continuous},
+    supports_weights=true,
+    reports_feature_importances=false,
+    load_path="SymbolicRegression.MLJInterfaceModule.SRRegressor",
+    descr="Symbolic Regression via Evolutionary Search",
```
OkonSamuel, line 232:
```suggestion
    human_name="Symbolic Regression via Evolutionary Search",
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,309 @@
...
+    Xnew_t = MMI.matrix(Xnew; transpose=true)
+    equations = params.equations
+    best_idx = params.best_idx
+    outs = [
+        let (out, flag) = eval_tree_array(eq[i], Xnew_t, fitresult.options)
+            !flag && error("Detected a NaN in evaluating expression.")
+            out
+        end for (i, eq) in zip(best_idx, equations)
+    ]
+    return reduce(hcat, outs)
```
OkonSamuel, line 177:
```suggestion
    outs_matrix = reduce(hcat, outs)
    prototype = MMI.istable(Xnew) ? Xnew : nothing
    return MMI.table(outs_matrix, prototype=prototype)
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,309 @@
...
+    params = MMI.fitted_params(m, fitresult)
+    Xnew_t = MMI.matrix(Xnew; transpose=true)
+    eq = params.equations[params.best_idx]
+    out, flag = eval_tree_array(eq, Xnew_t, fitresult.options)
+    !flag && error("Detected a NaN in evaluating expression.")
+    return out
+end
+function MMI.predict(m::MultitargetSRRegressor, fitresult, Xnew)
+    params = MMI.fitted_params(m, fitresult)
+    Xnew_t = MMI.matrix(Xnew; transpose=true)
```
OkonSamuel, line 168:
```suggestion
    Xnew_t, variable_names = get_matrix_and_colnames(Xnew)
```

src/MLJInterface.jl
OkonSamuel: predictions from multitarget regressor type models must be tables

src/MLJInterface.jl
MilesCranmer: Perhaps there should be a line here:

```julia
@assert variable_names == fitresult.variable_names
```

To prevent the user from accidentally passing a different table or ordering?

OkonSamuel: Yeah.

src/MLJInterface.jl
MilesCranmer: I'm trying this now. I think this assumes `X` has the same shape and variable names as `y`, no? 

Also, what if the user passes `X` as a regular matrix? Then, e.g.,

```julia
predict(mach, X) .- y
```

would throw an error, since `predict(mach, X)` is a table and `y` is a matrix.

OkonSamuel:
We are almost at the finish line

src/MLJInterface.jl
```diff
@@ -0,0 +1,309 @@
...
+    human_name="Multi-Target Symbolic Regression via Evolutionary Search",
+)
+
+function tag_with_docstring(model_name::Symbol, description::String)
+    docstring = """$(MMI.doc_header(eval(model_name)))
+
+    # Arguments
+    """
+
+    # TODO: These ones are copied (or written) manually:
```
OkonSamuel, line 250: We should be able to reuse some of the documentation from the main codebase. Right?

src/MLJInterface.jl
MilesCranmer: Can we store the schema of both `y` and `X` in `fitresult`, and use it as the `prototype`?

OkonSamuel: > I'm trying this now. I think this assumes `X` has the same shape and variable names as `y`, no?
> 
> Also, what if the user passes `X` as a regular matrix? Then, e.g.,
> 
> ```julia
> predict(mach, X) .- y
> ```
> 
> would throw an error, since `predict(mach, X)` is a table and `y` is a matrix.

We can work around that case

src/MLJInterface.jl
MilesCranmer: Yeah, most things are copied automatically in `OPTION_DESCRIPTIONS`. The remaining ones written out here are copied manually because the `equation_search` method has several arguments which are different from `SRRegressor`. But there should be a way to slice it cleanly in the future.

src/MLJInterface.jl
MilesCranmer: Check out what I just implemented. I think it solves both cases:

```julia
    X_t, variable_names, schema = get_matrix_and_colnames(X)
    y_t, y_variable_names, y_schema = format_input_for(m, y)
```

to get the column names and schema of both `y` and `X` (if necessary). Then, these are used in `predict` to give back a table of the same schema:

```julia
    outs = [
        let (out, flag) = eval_tree_array(eq[i], Xnew_t, fitresult.options)
            !flag && error("Detected a NaN in evaluating expression.")
            out
        end for (i, eq) in zip(best_idx, equations)
    ]
    outs_matrix = reduce(hcat, outs)
    fitresult.y_schema === nothing && return outs_matrix
    return MMI.table(
        outs_matrix; names=fitresult.y_variable_names, prototype=fitresult.y_schema
    )
```
This gives you a table if you train on a table for `y`, and a matrix if you train on a matrix.

OkonSamuel: prototype accepts a table source

OkonSamuel: We have to check in stages.

src/MLJInterface.jl
```diff
@@ -0,0 +1,324 @@
...
+    !flag && error("Detected a NaN in evaluating expression.")
+    return out
+end
+function MMI.predict(m::MultitargetSRRegressor, fitresult, Xnew)
+    params = MMI.fitted_params(m, fitresult)
+    Xnew_t, variable_names, schema = get_matrix_and_schema(Xnew)
+    @assert(
+        variable_names == fitresult.variable_names,
+        "Variable names do not match fitted regressor."
+    )
```
MilesCranmer, line 179: @OkonSamuel Should we also do a check for `schema == fitresult.schema`?

src/MLJInterface.jl
```diff
@@ -0,0 +1,324 @@
...
+    report = full_report(m, fitresult)
+    return (;
+        best_idx=report.best_idx,
+        equations=report.equations,
+        equation_strings=report.equation_strings,
+    )
+end
+function MMI.predict(m::SRRegressor, fitresult, Xnew)
+    params = MMI.fitted_params(m, fitresult)
+    Xnew_t = MMI.matrix(Xnew; transpose=true)
```
MilesCranmer, line 167: (Should we check `variable_names` here as well?)

src/MLJInterface.jl
OkonSamuel: No, there isn't any need for that.

src/MLJInterface.jl
```diff
@@ -0,0 +1,324 @@
...
+    equations = params.equations
+    best_idx = params.best_idx
+    outs = [
+        let (out, flag) = eval_tree_array(eq[i], Xnew_t, fitresult.options)
+            !flag && error("Detected a NaN in evaluating expression.")
+            out
+        end for (i, eq) in zip(best_idx, equations)
+    ]
+    outs_matrix = reduce(hcat, outs)
+    fitresult.y_schema === nothing && return outs_matrix
```
OkonSamuel, line 189: my point is `schema===nothing` doesn't necessarily mean that it must be a matrix or vector

src/MLJInterface.jl
MilesCranmer: > prototype accepts a table source not the schema

Should we store the original `y` input in `fitresult` so that we can use the correct prototype? Or are there any other lightweight things we could store?

src/MLJInterface.jl
OkonSamuel: There is a cleaner way to solve the issue. Could you revert the code first.

MilesCranmer: Reverted.

src/MLJInterface.jl
```diff
@@ -0,0 +1,318 @@
...
+        return_state=true,
+        loss_type=m.loss_type,
+    )
+    fitresult = (;
+        state=search_state,
+        options=options,
+        variable_names=variable_names,
+        y_variable_names=y_variable_names,
+    )
+    return (fitresult, nothing, full_report(m, fitresult))
```
OkonSamuel, line 127:
```suggestion
    return (fitresult, (nothing, MMI.istable(y)), full_report(m, fitresult))
```

OkonSamuel, line 127:
```suggestion
    return (fitresult, (nothing, MMI.istable(y)), full_report(m, fitresult))
```

src/MLJInterface.jl
MilesCranmer: Added.

src/MLJInterface.jl
OkonSamuel: So instead of storing the schema, we store the result of `MMI.istable(y)`

src/MLJInterface.jl
```diff
@@ -0,0 +1,322 @@
...
+
+"""Get an equivalent `Options()` object for a particular regressor."""
+function get_options(::AbstractSRRegressor) end
+
+eval(modelexpr(:SRRegressor))
+eval(modelexpr(:MultitargetSRRegressor))
+
+# Cleaning already taken care of by `Options` and `equation_search`
+function full_report(m::AbstractSRRegressor, fitresult)
+    _, hof = fitresult.state
```
OkonSamuel, line 74:
```suggestion
    _, hof = fitresult[1].state
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,322 @@
...
+end
+
+"""Get an equivalent `Options()` object for a particular regressor."""
+function get_options(::AbstractSRRegressor) end
+
+eval(modelexpr(:SRRegressor))
+eval(modelexpr(:MultitargetSRRegressor))
+
+# Cleaning already taken care of by `Options` and `equation_search`
+function full_report(m::AbstractSRRegressor, fitresult)
```
OkonSamuel, line 73:
```suggestion
function full_report(m::AbstractSRRegressor, (fitresult, _))
```

src/MLJInterface.jl
MilesCranmer: Why not add it to `fitresult` as an `:istable` key?

src/MLJInterface.jl
```diff
@@ -0,0 +1,322 @@
...
+
+function MMI.fitted_params(m::AbstractSRRegressor, fitresult)
+    report = full_report(m, fitresult)
+    return (;
+        best_idx=report.best_idx,
+        equations=report.equations,
+        equation_strings=report.equation_strings,
+    )
+end
+function MMI.predict(m::SRRegressor, fitresult, Xnew)
```
OkonSamuel, line 169:
```suggestion
function MMI.predict(m::SRRegressor, (fitresult, _), Xnew)
```

src/MLJInterface.jl
```diff
@@ -0,0 +1,322 @@
...
+        runtests=m.runtests,
+        saved_state=old_fitresult.state,
+        return_state=true,
+        loss_type=m.loss_type,
+    )
+    fitresult = (;
+        state=search_state,
+        options=options,
+        variable_names=variable_names,
+        y_variable_names=y_variable_names,
```
MilesCranmer, line 125:
```suggestion
        y_variable_names=y_variable_names,
        y_is_table=MMI.istable(y),
```
Perhaps this instead @OkonSamuel?

OkonSamuel: Yeah

src/MLJInterface.jl
OkonSamuel: You can also do that

MilesCranmer: Great, that works! Thanks!

MilesCranmer:
Okay the output format is updated and seems to work.

MilesCranmer:
Okay I have added the full docstrings. Anything else missing?

src/MLJInterface.jl
```diff
@@ -0,0 +1,488 @@
...
+            """
+    # Fitted parameters
+    - `equations::Vector{Node}`: The expressions discovered by the search, represented
+      in a dominating pareto frontier (i.e., the best expressions found for
+      each complexity).
+    - `equation_strings::Vector{String}`: The expressions discovered by the search,
+      represented as strings for easy inspection.
+    - `best_idx::Int`: The index of the best expression in the pareto frontier,
+       as determined by the `selection_method` function.
+
```
OkonSamuel, line 389: Can you add a report sections, since your report are non empty.
see https://github.com/JuliaAI/MLJDecisionTreeInterface.jl/blob/dev/src/MLJDecisionTreeInterface.jl#L640

src/MLJInterface.jl
```diff
@@ -0,0 +1,488 @@
...
+    # Fitted parameters
+    - `equations::Vector{Vector{Node}}`: The expressions discovered by the search, represented
+      in a dominating pareto frontier (i.e., the best expressions found for
+      each complexity). The outer vector is indexed by target variable, and the inner
+      vector is ordered by increasing complexity.
+    - `equation_strings::Vector{Vector{String}}`: The expressions discovered by the search,
+      represented as strings for easy inspection.
+    - `best_idx::Vector{Int}`: The index of the best expression in each pareto frontier,
+        as determined by the `selection_method` function.
+
```
OkonSamuel, line 464: Same here see above.

OkonSamuel (changes requested):
I can confirm that the documentation renders properly on my terminal.
This looks very good and detailed. Thanks @MilesCranmer, We can wait to have this addition to the MLJ registry.

MilesCranmer:
Added!

Once it merges, what should I do to add it to the MLJ registry? Should I submit an issue requesting a registry update on the MLJ repo?

MilesCranmer:
@OkonSamuel I also realized I need to fix the behavior of sample weights for multitarget case. What is the expected format of the sample weights `w` in multitarget problems? Is it expected that `w` is always a 1D array, or is `w` to be a 2D array, with columns corresponding to the same columns of `y`?

MilesCranmer:
Actually I think I just fixed it. MLJ assumes `w` is always a 1D vector, even for multitarget problems, correct? (This is what I expect; I just want to make sure it doesn't assume tabular sample weights).

OkonSamuel (approved):
LGTM!!!

MilesCranmer:
Thanks so much for your help @OkonSamuel; much appreciated!

PR #243, opened 2023-08-04
Evaluate on fixed batch when building per-pop hall of fame

MilesCranmer:
1. This change should greatly improve the effectiveness of batched losses, as the per-population hall of fame will not simply end up being the expression which got evaluated on the nicest batch.
2. This also removes the `fast_cycle` functionality to ease maintenance burden, as I don't think anybody was using it.

### DIFF
--- a/src/LossFunctions.jl
+++ b/src/LossFunctions.jl
@@ -1,6 +1,6 @@
 module LossFunctionsModule
 
-import Random: randperm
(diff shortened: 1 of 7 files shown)


MilesCranmer:
Would be good if this could depend on https://github.com/SymbolicML/DynamicExpressions.jl/pull/46 for faster hashing

PR #271, opened 2023-12-28
Program synthesis/graph expressions

MilesCranmer:
This is a working implementation of program synthesis in SymbolicRegression.jl via the use of nodes that are shared between multiple parents.

- Moving these TODOs to later:
    - Improving printing to be like a program, rather than like an expression
    - Consider optimizing the weight of forming connections and breaking connections
    - Consider whether we want to be smart at all about where we form connections across a tree
    - Create some unittests for these new mutations
    - Refactor existing mutations to be simpler (?)

- [x] Try other settings for `Optim.options`:
  - [x]  Set `f_calls_limit` in `Optim.options`, rather than just `iterations`.


MilesCranmer:
Debugging type instability from the `Node` type... Seems like Julia isn't specializing for some reason

MilesCranmer:
Did a git bisect on the illegal instruction error that comes up from the deterministic test. It started on 3a33882ed149817aee74ed577854c65d5b10d6a9

MilesCranmer:
The function `optimize_and_simplify_population` seems to be accounting for the performance regression here, even though `optimize_constants` by itself seems fine. So perhaps the simplification method is hurting us here?

MilesCranmer:
Need to set `f_calls_limit` in `Optim.options`.

MilesCranmer:
Seems like it's actually a performance regression in Optim.jl! So nothing caused by me.

PR #276, opened 2024-01-02
Implement swap operands for binary ops

foxtran:
For pow and for divide swapping of arguments may have some effect.

### DIFF
--- a/src/Mutate.jl
+++ b/src/Mutate.jl
@@ -13,6 +13,7 @@ import ..MutationFunctionsModule:
     gen_random_tree_fixed_size,
     mutate_constant,
     mutate_operator,
+    swap_operands,
     append_random_op,
     prepend_random_op,
     insert_random_op,
@@ -28,6 +29,7 @@ function condition_mutation_weights!(
         # If equation is too small, don't delete operators
         # or simplify
         weights.mutate_operator = 0.0
+        weights.swap_operands = 0.0
         weights.delete_node = 0.0
         weights.simplify = 0.0
         if !member.tree.constant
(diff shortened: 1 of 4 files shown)


src/Mutate.jl
```diff
@@ -37,6 +39,11 @@ function condition_mutation_weights!(
         return nothing
     end
 
+    if member.tree.degree != 2
```
MilesCranmer, line 42:
```suggestion
    if !any(node -> node.degree == 2, member.tree)
```

src/OptionsStruct.jl
```diff
@@ -40,6 +42,7 @@ will be normalized to sum to 1.0 after initialization.
 function MutationWeights(;
     mutate_constant=0.048,
     mutate_operator=0.47,
+    swap_operands=0.024,
```
MilesCranmer, line 45:
```suggestion
    swap_operands=0.0,
```
Just to follow SemVer, we can introduce this initially at 0.0 for a patch update (0.23.1), so that existing deterministic searches do not change, and then later we can adjust the weight to this. Does that sound good?

MilesCranmer (commented):
Nice work and great idea! I think this will help some searches quite a lot.

I left a couple of comments and then we can merge.

Cheers!
Miles

src/MutationFunctions.jl
```diff
@@ -29,6 +29,16 @@ function random_node(tree::Node{T}; filter::F=Returns(true))::Node{T} where {T,F
     return chosen_node[]
 end
 
+# Swap operands in binary operator for ops like pow and divide
+function swap_operands(tree::Node{T}, options::Options)::Node{T} where {T}
+    if !(has_operators(tree))
```
MilesCranmer, line 34:
```suggestion
    if !any(node -> node.degree == 2, tree)
```

MilesCranmer (approved):
Nice work! Thanks for being a contributor 🚀

PR #289, opened 2024-02-17
Make it easier to select expression from Pareto front for evaluation

MilesCranmer:
This enables the following syntax:

```julia
model = SRRegressor()
mach = machine(model, X, y)
fit!(mach)

# Predict with 3rd equation:
predict(mach, (data=X, idx=3))

# Predict with most complex equation:
r = report(mach)
predict(mach, (data=X, idx=lastindex(r.equations)))
```

which lets the user specify the equation they wish to use for prediction from the Pareto front.

For multiple outputs:

```julia
model = MultitargetSRRegressor()
mach = machine(model, X, y)
fit!(mach)


# Choose the 1st equation for output 1, 10th for output 2, and 5th for output 3:
predict(mach, (data=X, idx=[1, 10, 5]))
```

TODO:

- [x] Add example to documentation
- [x] Add example to README
- [x] Remove discussion of modifying `selection_method` (Seems to not work, see https://github.com/MilesCranmer/PySR/discussions/543 by @MrChewi)

---

@ablaom I would be interested to know if there is any way to make this sort of behavior compatible with MLJ? As it stands I think this might break some MLJ interfaces for users who wish to use this. (They would still be able to use `selection_method` parameter, but only before `fit!`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,16 +153,20 @@ predict(mach, X)
 ```
 
 This will make predictions using the expression
-selected using the function passed to `selection_method`.
-By default this selection is made a mix of accuracy and complexity.
-For example, we can make predictions using expression 2 with:
+selected by `model.selection_method`,
+which by default is a mix of accuracy and complexity.
+
+You can override this selection and select an equation from
+the Pareto front manually with:
 
 ```julia
-mach.model.selection_method = Returns(2)
-predict(mach, X)
+predict(mach, (data=X, idx=2))
 ```
 
-For fitting multiple outputs, one can use `MultitargetSRRegressor`.
+where here we choose to evaluate the second equation.
+
+For fitting multiple outputs, one can use `MultitargetSRRegressor`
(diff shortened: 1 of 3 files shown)


ablaom:
Sorry for not catching this earlier. This looks very breaking to me. In MLJ you cannot attach metadata to data at predict time.

ablaom:
Something similar was encountered in [hierarchical clustering](https://github.com/jbrea/MLJClusteringInterface.jl/blob/41d3c2195ad33f1840596c9762a3a67b9a124c6a/src/MLJClusteringInterface.jl#L614). In that case we wanted the user to choose, after fitting, the height of the dendrogram. There was quite a bit of discussion and the best we came up with was to return a *method* in the report to let the user predict with a user-specified height. (A hyperparameter controls the default height for the ordinary `predict`). 

I do have the idea to allow more flexible `predict` arguments in the LearnAPI.jl proposal, but I think it's difficult to justify in MLJ for small number of use-cases, as adding this is quite complicated.

PR #296, opened 2024-03-22
Better specialization

MilesCranmer:
PR to see if the lack of `OperatorEnum` specialization was what caused the 10% performance regression.


MilesCranmer:
Nope; doesn't seem to help at all...

PR #297, opened 2024-03-23
Ensure we load ClusterManagers.jl on workers

MilesCranmer:
This is a workaround for the dynamically created workers. It still seems like the Julia support is not great for a workflow with dynamic distributed computing, so this workaround seems pretty fragile.


MilesCranmer:
Cool, this seems to fix distributed computation for me!

PR #302, opened 2024-04-10
A small update on examples.md for 1-based indexing

liuyxpp:
Since we are in Julia not in Python, the indices of `x3` and `x0` should really be `x4` and `x1` to be consistent with the code to avoid any confusion.


MilesCranmer:
Good point! Thanks

PR #303, opened 2024-04-11
Fixes for Julia 1.11

MilesCranmer:
Julia 1.11 didn't work out-of-the-box so this PR fixes some issues with it and adds it to testing.

Seems like `Base.MainInclude.eval` was removed: https://github.com/JuliaLang/julia/issues/54057 so we can replace with `Core.eval(Base.MainInclude, ...)`. (Isn't this part of the public API though?)

### DIFF
--- a/src/Configure.jl
+++ b/src/Configure.jl
@@ -246,12 +246,7 @@ function import_module_on_workers(procs, filename::String, options::Options, ver
     else
         @info "Importing SymbolicRegression on workers as well as extensions $(join(relevant_extensions, ',' * ' '))."
     end
-    if VERSION < v"1.11.0-alpha1"
-        @everywhere procs Base.MainInclude.eval($expr)
-    else
-        # See https://github.com/JuliaLang/julia/issues/54057
-        @everywhere procs Core.eval(Base.MainInclude, $expr)
-    end
+    @everywhere procs Core.eval(Core.Main, $expr)
     return verbosity > 0 && @info "Finished!"
 end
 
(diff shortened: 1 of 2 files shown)


src/Configure.jl
```diff
@@ -246,7 +246,12 @@ function import_module_on_workers(procs, filename::String, options::Options, ver
     else
         @info "Importing SymbolicRegression on workers as well as extensions $(join(relevant_extensions, ',' * ' '))."
     end
-    @everywhere procs Base.MainInclude.eval($expr)
+    if VERSION < v"1.11.0-alpha1"
+        @everywhere procs Base.MainInclude.eval($expr)
+    else
+        # See https://github.com/JuliaLang/julia/issues/54057
+        @everywhere procs Core.eval(Base.MainInclude, $expr)
```
Keno, line 253:
```suggestion
        @everywhere procs Core.eval(Core.Main, $expr)
```

Keno: And this works unconditionally, so you don't need the version check

MilesCranmer: Cool, seems to work, thanks!

Btw, is this the right way to load extension code on dynamically allocated processes? I basically just:

1. Define a list of modules that trigger extensions, 
2. Check which ones are loaded on the head worker,
3. Load those modules on the workers

There are other difficulties in that the workers don't pick up the `v1.11` project and just activate directly in the current project, which might mean they don't have certain modules available. But I suppose that is unavoidable.

Keno: Don't know. The Distributed stuff predates our current dynamic loading implementation, so there isn't an intended way to do it per-se. If it works, I think it's fine.

MilesCranmer: Ok thanks, will do this for now

PR #305, opened 2024-04-25
Avoid `Base.sleep`

MilesCranmer:
Julia's implementation of `sleep` will sleep for more than a millisecond even if you request 1 microsecond.

**Edit:** Initial implementation of this PR switched to `Libc.systemsleep` inside a `Threads.@spawn`, but that felt a bit hacky as it relies on the Julia scheduler to actually start it.

Therefore this PR just removes `sleep` altogether. It will use a bit more CPU but it's necessary for highest performance.

x-ref https://github.com/JuliaLang/julia/issues/54971 – in the future if there's a high-resolution `sleep` we could probably switch back which would save some CPU


MilesCranmer:
Hm.. probably not worth the complexity..

MilesCranmer:
Okay I tested this on a machine with many CPUs, and it actually got a 30% performance improvement for the multi-threaded version. So I think it's worth it.

PR #326, opened 2024-06-24
BREAKING: Change expression types to `DynamicExpressions.Expression` (from `DynamicExpressions.Node`)

MilesCranmer:
These new experimental `Expression` types store both the operators and variable names within the object, rather than the plain `Node` which only stores the enum information about an expression.

This also adds `ParametricExpression` to learn basis expressions that have variable constants depending on class:

```julia
using SymbolicRegression
using Random: MersenneTwister
using Zygote
using MLJBase: machine, fit!, predict

rng = MersenneTwister(0)
X = NamedTuple{(:x1, :x2, :x3, :x4, :x5)}(ntuple(_ -> randn(rng, Float32, 30), Val(5)))
X = (; X..., classes=rand(rng, 1:2, 30))
p1 = rand(rng, Float32, 2)
p2 = rand(rng, Float32, 2)

y = [
    2 * cos(X.x4[i] + p1[X.classes[i]]) + X.x1[i]^2 - p2[X.classes[i]] for
    i in eachindex(X.classes)
]

model = SRRegressor(;
    niterations=10,
    unary_operators=[cos, exp],
    populations=10,
    expression_type=ParametricExpression,  # Subtype of `AbstractExpression`
    expression_options=(; max_parameters=2),
    autodiff_backend=:Zygote,
    parallelism=:multithreading,
)

mach = machine(model, X, y)
fit!(mach)
ypred = predict(mach, X)
```

so it basically learns $y= 2 \cos(x_4 + \alpha) + x_1^2 - \beta$ for $\alpha$ and $\beta$ parameters (which can be different according to the `classes` parameter – here there are two classes/types of behavior).

This `ParametricExpression` is just a single implementation of `AbstractExpression` but you can see how you can do pretty custom things now.

Fixes #340. Fixes #337. Fixes #336.

---

TODO:
- [x] Allow passing a `class` feature to MLJ which will have special treatment.
- [x] Debug why some of the tests seem to get stuck and take 3x longer to finish than normal.
- [x] Consider documenting this, or just leaving it as an experimental undocumented feature until it stabilizes.
- [x] Add Enzyme backend.
- [x] Add example to docs.
- [x] Consider moving to Literate.jl for docs?
- [x] Fix ResourceMonitor weirdness


ext/SymbolicRegressionSymbolicUtilsExt.jl
```diff
@@ -11,10 +12,14 @@ import SymbolicRegression: node_to_symbolic, symbolic_to_node
 
 Convert an expression to SymbolicUtils.jl form.
 """
-function node_to_symbolic(tree::AbstractExpressionNode, options::Options; kws...)
-    return node_to_symbolic(tree, options.operators; kws...)
+function node_to_symbolic(
+    tree::Union{AbstractExpressionNode,AbstractExpression}, options::Options; kws...
+)
+    return node_to_symbolic(get_tree(tree), options.operators; kws...)
```
MilesCranmer, line 18: Should get operators that were passed in (if)

ext/SymbolicRegressionSymbolicUtilsExt.jl
```diff
@@ -31,24 +36,30 @@ function symbolic_to_node(eqn::Symbolic, m::AbstractSRRegressor; kws...)
...
 
 function Base.convert(
-    ::Type{Symbolic}, tree::AbstractExpressionNode, options::Options; kws...
+    ::Type{Symbolic},
+    tree::Union{AbstractExpressionNode,AbstractExpression},
+    options::Options;
+    kws...,
 )
-    return convert(Symbolic, tree, options.operators; kws...)
+    return convert(Symbolic, get_tree(tree), options.operators; kws...)
```
MilesCranmer, line 44: Same issue

src/Options.jl
```diff
@@ -723,14 +736,23 @@ $(OPTION_DESCRIPTIONS)
 
     @assert print_precision > 0
 
+    _autodiff_backend = if autodiff_backend isa Union{Nothing,AbstractADType}
+        autodiff_backend
+    else
+        ADTypes.Auto(autodiff_backend)
```
MilesCranmer, line 742: @gdalle here's the usecase^ thanks again :+1:

gdalle: awesome! do you want me to give a quick review?

MilesCranmer: It’s a pretty minor part of the PR so not needed yet. But once I try to get more backends working in the future it would be much appreciated, thanks! :)

gdalle: alright, don't hesitate to ping me as needed!

MilesCranmer:
I ran the benchmark on my machine and don't see any performance differences. Not sure what's up with the github action version...

MilesCranmer:
Weirdly it seems like the gradients are never actually being used in the Zygote optimisation... They are always `nothing` and I didn't notice it. I guess Optim.jl automatically switches to finite difference if the gradient errors.

MilesCranmer:
@gdalle would you happen to know how to force custom ChainRules rrule/frule onto other backends? I think part of the reason they may not work is because `eval_tree_array` is too complicated for them to trace. But if they just use the rrule/frule they should be good.

gdalle:
> They are always nothing and I didn't notice it. I guess Optim.jl automatically switches to finite difference if the gradient errors.

I would be very surprised if it did that. This kind of trick is familiar within the SciML ecosystem but (hopefully) not widespread outside ^^

> @gdalle would you happen to know how to force custom ChainRules rrule/frule onto other backends?

What do you mean by that? Which backends, and what do you want to "force"?

MilesCranmer:
I remember reading this section on ChainRules.jl: https://juliadiff.org/ChainRulesCore.jl/dev/index.html#Packages-supporting-importing-rules-from-ChainRules.

> Several packages do not automatically load rules from ChainRules by default, but support importing rules that were defined using it, e.g. with a macro.

I guess I am wondering if there is a multi-backend way to trigger this "import" to happen, perhaps from within DifferentiationInterface.jl. 

I have an rrule defined here: https://github.com/SymbolicML/DynamicExpressions.jl/blob/9e95f0538207c360874615686be5c0ec627aee42/src/ChainRules.jl#L22 and am looking to import it across different backends.

gdalle:
No there's nothing like that at the moment. The reverse functionality exists though, as [`DifferentiateWith`](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/stable/api/#DifferentiationInterface.DifferentiateWith) (define a chain rule from another backend).

Which backends are you hoping to use other than Zygote.jl?

MilesCranmer:
> No there's nothing like that at the moment. The reverse functionality exists though, as [`DifferentiateWith`](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/stable/api/#DifferentiationInterface.DifferentiateWith) (define a chain rule from another backend).

Cool!

> Which backends are you hoping to use other than Zygote.jl?

Enzyme.jl and ForwardDiff.jl (ForwardDiff seems to work, but I'd like it to use my `frule` as its faster than the trace-generated version)

gdalle:
> Enzyme.jl and ForwardDiff.jl (ForwardDiff seems to work, but I'd like it to use my frule as its faster than the default traced version)

Note that neither of these backends is mentioned in https://juliadiff.org/ChainRulesCore.jl/dev/index.html#Packages-that-automatically-load-rules-from-ChainRules, hence my question.
As I explain [here](https://discourse.julialang.org/t/ad-pipeline-and-hessian-vector-products/116329/4?u=gdalle), ForwardDiff and Enzyme both have their own sets of rules.

- To import a ChainRule into ForwardDiff, use https://github.com/ThummeTo/ForwardDiffChainRules.jl
- To import a ChainRule into Enzyme, use https://github.com/EnzymeAD/Enzyme.jl/pull/996 (not sure it's part of the public API)

gdalle:
Sorry, I was looking at the section above, packages that load automatically. My mistake.
You're right, both ForwardDiff and Enzyme support manual loading, and the tools necessary are the ones I linked to.
I will have to think about whether DI should support these translation utilities. But I think it's really hard cause each rule system has its own syntax, and metaprogramming makes the whole thing even harder. So feel free to open an issue in DI but it might be low on my list of priorities.

gdalle:
Basically, supporting rule translation in DI would involve defining a universal rule definition syntax. The universal differentiation syntax was already tough to iron out, but the universal rule definition syntax is another level entirely.

MilesCranmer:
I think the parametric expressions might be getting their metadata (incl. parameters) stripped when a model is updated?

MilesCranmer (commented):
Reviewed everything again, seems good once the tests all pass

src/Options.jl
```diff
@@ -28,65 +28,111 @@ using ..OperatorsModule:
...
     end
-    if typeof(una_constraints) <: Array && !is_una_constraints_already_done
-        una_constraints = Dict(una_constraints)
+    _una_constraints2 = if _una_constraints1 === nothing
+        fill(-1, length(unary_operators))
+    elseif !is_una_constraints_already_done
+        [
+            haskey(_una_constraints1, op) ? _una_constraints1[op]::Int : -1 for
+            op in unary_operators
+        ]
```
atharvas, line 56: You'll need an `else una_constraints1` condition here, otherwise `_una_constraints2` will be `nothing` when the arguments are valid.

Same for `bin_constraints2`. It should look like:

```julia
@unstable function build_constraints(;
    una_constraints,
    bin_constraints,
    @nospecialize(unary_operators),
    @nospecialize(binary_operators)
)::Tuple{Vector{Int},Vector{Tuple{Int,Int}}}
    # Expect format ((*)=>(-1, 3)), etc.
    # TODO: Need to disable simplification if (*, -, +, /) are constrained?
    #  Or, just quit simplification is constraints violated.

    is_una_constraints_already_done = una_constraints isa Vector{Int}
    _una_constraints1 = if una_constraints isa Array && !is_una_constraints_already_done
        Dict(una_constraints)
    else
        una_constraints
    end
    _una_constraints2 = if _una_constraints1 === nothing
        fill(-1, length(unary_operators))
    elseif !is_una_constraints_already_done
        [
            haskey(_una_constraints1, op) ? _una_constraints1[op]::Int : -1 for
            op in unary_operators
        ]
    else
        una_constraints1
    end

    is_bin_constraints_already_done = bin_constraints isa Vector{Tuple{Int,Int}}
    _bin_constraints1 = if bin_constraints isa Array && !is_bin_constraints_already_done
        Dict(bin_constraints)
    else
        bin_constraints
    end
    _bin_constraints2 = if _bin_constraints1 === nothing
        fill((-1, -1), length(binary_operators))
    elseif !is_bin_constraints_already_done
        [
            if haskey(_bin_constraints1, op)
                _bin_constraints1[op]::Tuple{Int,Int}
            else
                (-1, -1)
            end for op in binary_operators
        ]
    else
        bin_constraints2
    end

    return _una_constraints2, _bin_constraints2
end
```

atharvas:
I was encountering some issues with constraint parsing in `Options.jl`. Check out the comment. Don't know why the test cases don't catch the issue.

src/Options.jl
MilesCranmer: Thanks!

MilesCranmer:
Going to punt `StructuredExpression`s until later. @eelregit let me know if you are at all interested in this! `StructuredExpression` would let you evolve within a fixed functional form. Seems like there are a couple missing methods that would allow it to work but hopefully won't take too much work. I'll have to pause on this side of things for now.

MilesCranmer:
Seems like the garbage collection is going crazy in the tests, which is why they are so slow. The reason why 1.6 and 1.8 are much faster is – I think – because DispatchDoctor.jl is turned off. So something about DispatchDoctor.jl is causing the GC to overwork itself... Possibly related to https://github.com/MilesCranmer/DispatchDoctor.jl/issues/57 and https://github.com/MilesCranmer/DispatchDoctor.jl/issues/58?

MilesCranmer:
Fixed the performance regression in the unittests with https://github.com/SymbolicML/DynamicExpressions.jl/pull/102/commits/74c8dc1db5a0b192c49a35d7a8b1d8e5a792cd61.

PR #348, opened 2024-10-06
Implement tree rotation operator

MilesCranmer:
This implements a tree rotation as one of the mutation operators:

Though the code implementation here also covers unary operators – both as a root and as a pivot (and as both).

As an example, the expression $$1.5 \cdot \exp(x_1) + \frac{2.5}{x_3}$$ can randomly mutate into one of: $$\frac{(1.5 \cdot \exp(x_1)) + 2.5}{x_3},$$ $$1.5 \cdot \left(\exp(x_1) + \frac{2.5}{x_3}\right),$$ or $$\exp(1.5 \cdot x_1) + \frac{2.5}{x_3},$$.

Thanks to @larsentom for the idea.

~~This current implementation is a draft.~~ From watching the learning dynamics it definitely seems to help diversify the expressions (!) but we likely want to tune the default mutation weight.

- [x] It would also be nice to have `NodeSampler` be able to return parent nodes. Otherwise we either need to skip root nodes or do what I do currently which is do a second tree search for the parent - which seems inefficient. But maybe it is the best strategy, I'm not sure.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,7 +21,7 @@ jobs:
   test:
     name: Julia ${{ matrix.julia-version }}-${{ matrix.os }}-${{ matrix.test }}-${{ github.event_name }}
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 120
+    timeout-minutes: 240
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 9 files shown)


MilesCranmer (commented):
Going to modify this so it works for unary operators too

MilesCranmer:
cc @foxtran – seems this new operator complements your swap operator well

foxtran:
Nice!

PR #353, opened 2024-10-12
Create overloadable utilities: `AbstractOptions`, `AbstractRuntimeOptions`, `AbstractMutationWeights`, `AbstractSearchState`, and `mutate!`

MilesCranmer:
@atharvas this should help simplify LaSR and make it compatible with future versions of SymbolicRegression.jl

Basically the trick is to do the following:

```julia
struct LaSROptions{O<:SR.Options} <: SR.AbstractOptions
    lm_options::LLMOptions
    sr_options::O
end
const LLM_OPTIONS_KEYS = fieldnames(LLMOptions)
@inline function Base.getproperty(options::LaSROptions, k)
    if k in LLM_OPTIONS_KEYS
        return getproperty(getfield(options, :lm_options), k)
    else
        return getproperty(getfield(options, :sr_options), k)
    end
end
```

Then you can pass any instance of `LaSROptions` directly to `equation_search`.

Furthermore, it lets you overload the behavior of any function inside SymbolicRegression.jl that has an input argument `options` – simply declare a new method for `::LaSROptions`.

What do you think?

You can also create a function `LaSROptions` that checks which keys are in `LLMOptions`, and uses those to instantiate it, and the other ones to instantiate `SR.Options`.


src/SymbolicRegression.jl
```diff
@@ -1184,7 +1100,7 @@ using ConstructionBase: ConstructionBase as _
 include("precompile.jl")
 redirect_stdout(devnull) do
     redirect_stderr(devnull) do
-        do_precompilation(Val(:precompile))
+        # do_precompilation(Val(:precompile))
```
MilesCranmer, line 1103:
```suggestion
        do_precompilation(Val(:precompile))
```

src/Mutate.jl
```diff
@@ -148,133 +206,43 @@ function next_generation(
...
-                    beforeLoss,
-                    options,
-                    compute_complexity(member, options);
-                    parent=parent_ref,
-                    deterministic=options.deterministic,
-                ),
-                mutation_accepted,
-                num_evals,
+        # TODO: This uses dynamic dispatch. But it doesn't seem that bad
+        # in terms of performance. Still should investigate in more detail.
```
MilesCranmer, line 211:
```suggestion
```

MilesCranmer:
@atharvas let me know if there’s anything else missing for LaSR and I can add it in another PR.

PR #355, opened 2024-10-13
Create `TemplateExpression` for providing a pre-defined functional structure and constraints

MilesCranmer:
# `TemplateExpression`

This lets you easily create expressions with a specific structure.

For example, if you want to have $$f(x_1, x_2) + exp(- g(x_3) \cdot g(x_3))$$, this PR will let you.

~~There are some remaining issues though that we need to fix. It seems there is an undefined reference error when simplifying that I think is likely due to `get_tree` stitching the inner expressions together, whereby some function (that I haven't figured out yet) mutates the resultant expressions.~~

~~This should never happen, and is always avoided by `mutate!` because of the fact that they call the new function `get_contents_for_mutation` which prevents this. But I guess there are some other functions which also need this guard in place.~~

# Fixes aliasing issues

This was a very bad bug that I just happened to identify as part of this PR ([[1]](https://github.com/MilesCranmer/SymbolicRegression.jl/pull/355#issuecomment-2409377179),[[2]](https://github.com/MilesCranmer/SymbolicRegression.jl/pull/355#issuecomment-2409415585)). Basically, the crossover operation would sometimes happen between a population member and **itself**. There was no check to prevent the crossover from sampling the exact same member from the population!

In addition to weird effects from actually performing a crossover operation on the same expression, this also meant that multiple individuals in a population could end up being the same. This means that all downstream operations (including ones ran in parallel!) would be susceptible to aliasing issues.

In short, this was a nasty bug, and it is surprising that it only showed up explicitly when implementing the BlueprintExpression. I wonder if this was the source of various crashes (especially for long-running jobs) over the past 1+ years, and I was just never able to identify the source of the race condition.

Anyways, this seems like it is fixed now, because we copy population members immediately after sampling.

# Fix `maxsize` issues

There were some issues hanging around that had to do with `maxsize` being violated. It turns out this was simply an issue of a `<` appearing where a `<=` should have been. This has been fixed.

---

TODO:

- [x] Fix aliasing issue due to some function mutating the combined expression rather than working only on a single inner expression at a time.
  - One idea is that I could create a new node type, `ReadOnlyNode`, which throws an error for any `setproperty!`. Calling `get_tree` on `ConstrainedExpression` will return this node type.
    - This was implemented in https://github.com/SymbolicML/DynamicExpressions.jl/pull/105
- [x] Fix a name for this. Currently it is `ConstrainedExpression` but I'm not a fan of this. `ConstrainedStructuredExpression` is too long. Maybe `MultiExpression`?
- [x] Document the new type.
- [x] Add unittests. Want to reach 100% diff coverage as there's a lot of new features here.


MilesCranmer:
I wonder if the aliasing could be due to the use of `copy_node` rather than `copy` in the codebase? Note that `copy` calls `copy_node` but not the other way around.

MilesCranmer:
It looks like sometimes, crossover attempts to cross the same expression with itself. Perhaps that is why this issue arises? Not because of the fact that there is aliasing, but because `copy` corrects any undefined fields in the tree?

MilesCranmer:
`TemplateExpression` might be better actually

PR #360, opened 2024-10-20
Output folder, better TemplateExpression, colored printouts, switch to ProgressMeter

MilesCranmer:
- [x] This creates a full CHANGELOG.md based on all of the release notes. It also creates the 1.0.0 release notes giving descriptions of the various new features.
- [x] Make TemplateExpression more robust
- [x] Nice example with structs
- [x] ~~New URL~~ Later.
- [x] Make a proper output folder, rather than dumping halls of fame to the current dir
- [x] ~~Permit constraints and simplification by default~~ Will add later if needed.
- [x] Bump to DispatchDoctor.jl 0.4.17 for `@nospecialize` fix
- [x] ~~Set up user mailing list – send around an email that people can accept to join the google group for PySR/SymbolicRegression.jl updates~~ Later.
- [x] ~~Custom complexities, custom constraints, etc for TemplateExpression~~ Will add later if needed.
- [x] ~~Better default printout for TemplateExpressions - should print individual components!~~ Cancelled this for now.


MilesCranmer:
I think in the example it would be nice to have an example with a literal user-defined struct getting returned from the structure function. And the fields of that struct could depend nonlinearly on the functions evolved.

MilesCranmer:
e.g.,

```
struct Energy{T}
    hamiltonian::T
    lagrangian::T
end

energy(nt) = map(
    (T, V) -> Energy(T + V, T - V),
    nt.T, nt.V
)
energy(nt::NamedTuple{<:Any,Tuple{Vararg{String}}}) = (
    "(T = $(nt.T), V = $(nt.V))"
)

variable_mapping = (; T = [1], V = [2])
```

and then do a search based on some values of H and L!

PR #377, opened 2024-12-07
Expose recorder function for DynamicAutodiff.jl

MilesCranmer:
This makes it so you can put derivative operators in the template structure when doing structured searches. This works with a derivative operator `D` that computes the derivative with respect to the `i`-th argument, and can be nested arbitrarily to get higher-order derivatives. However, note that nesting it multiple times could be take a while to compile though, since it will compile n^d operators for d-th order derivatives. (Runtime performance should be unaffected)

Here's an example:

```julia
using SymbolicRegression
using SymbolicRegression: D

structure = TemplateStructure{(:f, :g)}(
    ((; f, g), (x, y)) -> f(x, y) - D(f, 1)(x, y) + g(y)
)
```

This imposes a structure that for some symbolic $f$ and $g$, the evaluation would use $f(x, y) - \frac{\partial f(x, y)}{\partial x} + g(y)$.

Also helps support feature request from @gm89uk + @DeaglanBartlett. Interested to hear whether you run into any issues for it before I merge it!


MilesCranmer:
Turns out Zygote.jl is awful at second-order differentiation, even for tiny operators like `cos`. So I will switch things to ForwardDiff.jl

Moelf:
is there any reason to use Enzyme or Mooncake for this?

MilesCranmer:
It is probably overkill as all I'm doing is:

1. Generating a derivative operator for each current operator
2. Creating a new operator enum with those derivative operators (inferrable, even with runtime-generated expressions)
3. Applying the chain rule and "shifting" the operator indices in the tree when we need to call the derivatives
4. Running regular `eval_tree_array` with that new operator enum

So all I'm using `ForwardDiff` for is this:

```julia
function (d::OperatorDerivative{F,1,1})(x) where {F}
    return ForwardDiff.derivative(d.op, x)
end
function (d::OperatorDerivative{F,2,1})(x, y) where {F}
    return ForwardDiff.derivative(Fix{2}(d.op, y), x)
end
function (d::OperatorDerivative{F,2,2})(x, y) where {F}
    return ForwardDiff.derivative(Fix{1}(d.op, x), y)
end
```

I've had a lot of issues in getting Enzyme into SR (https://github.com/EnzymeAD/Enzyme.jl/issues?q=is:issue%20author:MilesCranmer – though have appreciated the support from the authors) so want to avoid making it a direct dependencies until it is more stable. Mooncake I haven't tried yet but the README says it is still under development so probably want to avoid.

On the other hand, ForwardDiff.jl is already an indirect dependency via Optim.jl so I feel pretty safe about using it. What do you think?

src/ComposableExpression.jl
```diff
@@ -244,4 +248,264 @@ for op in (
...
+            if g_prime.degree == 0 && g_prime.constant && isone(g_prime.val)
+                return f_prime
+            else
+                return constructorof(N)(; op=ctx.mult_idx, l=f_prime, r=g_prime)
+            end
+        end
+    else  # tree.degree == 2
+        # f(g(x), h(x)) => f^(1,0)(g(x), h(x)) * g'(x) + f^(0,1)(g(x), h(x)) * h'(x)
+        f_prime_left_op = tree.op + ctx.nbin
+        f_prime_right_op = tree.op + 2 * ctx.nbin
```
MilesCranmer, line 333: @Moelf this is where the binary operator indices are shifted. It's nice because it basically lets us generate a symbolic derivative at runtime by literally just adding a couple integers.

Moelf:
>On the other hand, ForwardDiff.jl is already an indirect dependency via Optim.jl so I feel pretty safe about using it. What do you think?

that's totally true, yea

MilesCranmer:
Update: since the derivatives are general to any DynamicExpressions.jl expression, I'm going to register a separate package DynamicAutodiff.jl, then import that to SymbolicRegression.jl

PR #382, opened 2024-12-09
feat: allow user-specified `stdin`

### LINKED ISSUE #370: [BUG]: frozen REPL in VSCode
### What happened?

All the details can be found here:
https://discourse.julialang.org/t/symbolicregression-jl-freezing-my-terminal-when-running-report/119368/2

In summary, when using SymbolicRegression.jl in VS Code in windows, after running `fit!(match)` the REPL freezes and I can't type any more. It is non-responsive. 
I have been working with a local fork of the package modifying a line as recommended here: https://discourse.julialang.org/t/symbolicregression-jl-freezing-my-terminal-when-running-report/119368/5?u=elcup

### Version

v1.0.1

### Operating System

Windows

### Interface

Other (specify below)

### Relevant log output

_No response_

### Extra Info

Working on VS Code in Windows

MilesCranmer:
This lets the user specify a specific `IO` object to read user input from, which by default is `stdin`, but could also be some other stream.

This fixes #370.

@rafaelcuperman sorry you haven't received a response from the VSCode extension team, I think they are just pretty busy. In the meantime I hope this helps with your issue! You would just specify `input_stream=devnull` when running on your system and it should avoid the hang.

### DIFF
--- a/src/Options.jl
+++ b/src/Options.jl
@@ -868,7 +868,6 @@ $(OPTION_DESCRIPTIONS)
         deprecated_return_state::Union{Bool,Nothing},
         typeof(_autodiff_backend),
         print_precision,
-        typeof(input_stream),
     }(
         operators,
         _bin_constraints,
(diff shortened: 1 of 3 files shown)


rafaelcuperman:
Thank you very much for this. Works like a charm!

PR #385, opened 2024-12-10
refactor: reduce allocations from type assertion

MilesCranmer:


MilesCranmer:
Error looks to just be an ambiguity. I’ll update the TemplateExpression trait to be more generic

PR #387, opened 2024-12-12
Re-use allocations in mutation loop

MilesCranmer:
This re-uses allocations in the mutation loop. This should hopefully result in a nice speed up from reduced need for garbage collection.


MilesCranmer:
Hm. There’s no speedup… not sure it’s worth the added complexity?

PR #388, opened 2024-12-14
feat: add safe versions of `asin` and `acos`

MilesCranmer:


MilesCranmer:
Need to put in Options as well

PR #392, opened 2024-12-21
Broaden MLJ `target_scitype` only when using `TemplateExpression` 

MilesCranmer:
Fixes #390 

@ablaom how is this?

PR #393, opened 2024-12-23
fix: make `:class` col more generic to `X` type

MilesCranmer:
@ablaom this should fix the issue you noticed with `rowtable(X)` types.

However this will currently result in type instability for NamedTuple input since I have to `collect` the columns due to this issue: https://github.com/JuliaAI/MLJModelInterface.jl/issues/214. But it does fix the issue for other types.

### DIFF
--- a/src/InterfaceDynamicExpressions.jl
+++ b/src/InterfaceDynamicExpressions.jl
@@ -359,4 +359,7 @@ function DE.EvaluationHelpersModule._grad_evaluator(
     )
 end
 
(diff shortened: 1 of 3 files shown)


ablaom:
I've asked someone to take a look at the MLJModelInterface issue, which doesn't look too difficult to address. I'll leave it up to you whether you proceed with this PR or not. This isn't urgent from my point of view. If you would like me test if this resolves the issue at #392 then I'll need a PR combining both changes please (eg. merge this and rebase #392). Ping me when you 're ready.

PR #395, opened 2024-12-26
Change `get_tournament_selection_weights` function signature

atharvas:
`get_tournament_selection_weights`, presently, only accepts `options::Options`. This created a bottleneck for LaSR (or any other SymbolicRegression.jl customization) that relies on multiple dispatch of subtypes of `AbstractOptions` type. This commit resolves the issue by changing the function signature from `get_tournament_selection_weights(::Options)` to `get_tournament_selection_weights(::AbstractOptions)`. Changing this function signature has no/negligible impact on SymbolicRegression.jl test case performance.


atharvas:
This seems to be the only SR.jl-side code change needed to fully rewrite LaSR as a customization of SR.jl. 

Caveat: While the LaSR rewrite is complete, I'm progressively increasing LaSR's test coverage. This might uncover more SR.jl-side changes. I'm thinking of opening up a separate PR for other changes, but happy to keep this PR open and stack up all the changes in this PR.

src/Population.jl
```diff
@@ -167,7 +167,7 @@ const CACHED_WEIGHTS =
         PerThreadCache{Dict{Tuple{Int,Float32},typeof(test_weights)}}()
     end
 
-@unstable function get_tournament_selection_weights(@nospecialize(options::Options))
+@unstable function get_tournament_selection_weights(@nospecialize(options::AbstractOptions))
     n = options.tournament_selection_n
     p = options.tournament_selection_p
```
MilesCranmer, line 172:
```suggestion
    n = options.tournament_selection_n::Int
    p = options.tournament_selection_p::Float32
```

But it likely doesn't affect things that much. I think the original reason I limited this to `Options` rather than `AbstractOptions` was because having the `@nospecialize` would result in the compiler being unable to infer the type of each of these. But just having these assertions there will probably help.

MilesCranmer (approved):
Looks good!

PR #398, opened 2024-12-30
Add abstract types for single- and multi-target MLJ Regressors.

atharvas:
Right now, SRRegressor and MultitargetSRRegressor subtype AbstractSRRegresor. However, much of the `src/MLJInterface.jl` code dispatches on either SRRegressor or MultitargetSRRegressor. This creates a bottleneck for defining custom MLJInterfaces and forces redefinition of much of `src/MLJInterface.jl`. The commit addresses this by introducing two new types, SingletargetAbstractSRRegressor and MultitargetAbstractSRRegressor, and changes the logic in `src/MLJInterface.jl` to dispatch on these types instead.

Note: This approach is kind of janky as it solves the reuse problem using multiple inheritance (which might cause problems of its own if abused). This can be trimmed further / I'm open to rethinking this.


MilesCranmer:
Sounds good to me. One minor thing is it should be `AbstractMultitargetSRRegressor` per usual Julia style of `Abstract` going at the start of the type name. So perhaps `AbstractSRRegressor` should be named to `AbstractSymbolicRegressor` or something, so it sounds more "general"?

atharvas:
@MilesCranmer might need your help debugging this. The test case for parametric expressions with the Zygote autodiff seem to be failing even on the SymbolicRegression master branch.

MilesCranmer:
Sorry for the delay. This turned out to be from an update to DifferentiationInterface.jl which has since been fixed: 

- https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/713
- https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/714

Just waiting on a release now.

MilesCranmer (approved):
Looks good to me, thanks!!

PR #399, opened 2025-01-01
fix for `turbo` and `bumper` not being used in TemplateExpression

MilesCranmer:
cc @gm89uk - should fix the speed issue you noted over in the pysr discussions page


gm89uk:
Thanks @MilesCranmer for fixing this so quickly. I will do a long run overnight with this and report back. 
I'm looking forward to the next merge to run parametric template expressions with bumper and turbo!

Presumably no change in code is required here, just to turn bumper and turbo to 
Thanks again

gm89uk:
Thanks @MilesCranmer, I tested it out and it is quite a bit quicker than before! TemplateExpression is remains quite a bit slower after a long run than a custom loss function. 

Further thoughts:
_With monotonicity check_
TemplateExpression is definitely quicker than the loss_fnc to begin with, but then slows down (although not quite as much as before) 

What I think is going on, 
_Custom Loss Function_: `eval_grad_tree_array` is calculating grad for all variables (+prediction) in a single tree traversal (I think) and using Zygote for the ForwardPass derivative 

_TemplateExpression Code_:   o = f(..) is performing an `eval_tree_array`, then following a very fast symbolic differentiation, it is then repeating `eval_tree_array` for first derivative for each variable. 
So initially, the code might be doing one `eval_tree_array`, quickly failing the monotonicity check and exiting early for a faster runtime. Then, as the expressions evolve toward those that have monotonicity for variables 1-3, the full structure code is being processed and looping `eval_tree_array` a further 3 times.

Therefore, you are comparing the overhead of `eval_grad_tree_array` once in the custom loss function , vs `eval_tree_array` four times in the TemplateExpression, (one for  o = f(…) and then one for each variable in the monotonicity check loop with d(...).x) at later runs.

In summary for first diff with multiple variables, `eval_grad_tree_array` is probably more efficient than my structure code and I cannot think of a faster way to do those within a TemplateExpression, which is fine for most cases. It's just my database is relatively large at 7000+ rows (with batching set to 50).
 
However, interestingly, I ran a stripped down code without any monotonicity checks and found the custom_loss_func to run about twice as fast as TemplateExpression:

 Running stripped down code: 
**Template Expression**
```julia
using Statistics #for mean
structure = TemplateStructure{(:f,)}(
  ((; f), (x1, x2, x3, x4, x5, x6)) -> begin
    o = f(x1, x2, x3, x4, x5, x6)
    return o
  end
)
model = SRRegressor(
    niterations=1000000,
    binary_operators=[+,-,*,/],
    maxsize=60,
    bumper=true,
    turbo=true,
    populations=18,
    expression_type = TemplateExpression,
    expression_options = (; structure),
    population_size=100,
    parsimony = 0.01,
    batching=true,
)
```
**Custom Loss Function**
```julia
using Statistics #for mean

function loss_fnc(tree, dataset::Dataset{T,L}, options, idx) where {T,L}    
    # Extract data for the given indices
    X = idx === nothing ? dataset.X : view(dataset.X, : , idx)
    prediction, grad, complete = eval_grad_tree_array(tree, X, options; variable=true)
    if !complete
        return L(Inf)
    end
    return mean((prediction .- y).^2)
end
model = SRRegressor(
    niterations=1000000,
    binary_operators=[+,-,*,/],
    maxsize=60,
    bumper=true,
    turbo=true,
    populations=18,
    population_size=100
    parsimony = 0.01,
    batching=true,
    loss_function = loss_fnc,
)
```

So there is understandably some overhead from setting up a TemplateExpression structure there as well.

PR #404, opened 2025-01-31
fix: hotfix Zygote backend of DifferentiationInterface

MilesCranmer:
DifferentiationInterface introduced a change in behavior where the Zygote backend returning `nothing` is now an error rather than `nothing`; this implements a hotfix. It's not a great solution but I'd prefer this over another extension.

FYI @gdalle I think the ZygoteNothingError should have been introduced in a breaking release as normally it's common to handle `nothing` explicitly. ~~Though I do think it's the right behavior though overall.~~ Actually changed my mind, I think the `nothing` should just be returned.

Also would be nice to have ZygoteNothingError defined inside the package itself. If not wanting to put it there, perhaps you could have some trait like:

```julia
nothing_error_type(_) = Nothing


# in Zygote extension:
nothing_error_type(::AutoZygote) = ZygoteNothingError
```

then I can simply call `e isa nothing_error_type(backend)`.

Or maybe have an abstract type of error for when a gradient isn't defined?

### DIFF
--- a/src/ConstantOptimization.jl
+++ b/src/ConstantOptimization.jl
@@ -2,7 +2,7 @@ module ConstantOptimizationModule
 
 using LineSearches: LineSearches
 using Optim: Optim
-using ADTypes: AbstractADType, AutoEnzyme
+using ADTypes: AbstractADType, AutoEnzyme, AutoZygote
 using DifferentiationInterface: value_and_gradient
 using DynamicExpressions:
     AbstractExpression,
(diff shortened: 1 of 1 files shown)


MilesCranmer:
Fixed at the source with https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/714

gdalle:
https://github.com/JuliaRegistries/General/pull/124177

MilesCranmer:
Thanks!

PR #421, opened 2025-02-16
Change behavior of batching to resample only every iteration; not every eval 

### LINKED ISSUE #418: Batch should be chosen per iteration
I think that the batch should be chosen once per iteration, rather than per evaluation. Choosing it once per evaluation means that it will basically just end up picking the lucky batch.

MilesCranmer:
Fixes #418.

This refactors the dataset struct to be an abstract type, with two concrete classes: `BasicDataset` and `SubDataset`. This is cleaner than the current mechanism of passing `idx` around. Instead the `SubDataset <: Dataset` stores these internally, and `.X` returns a sliced version of the underlying `X` (same for `.y` and `.weights`, but not for other properties; though `.n` returns the batch size). 

This should be entirely backwards compatible, because `::Dataset` will still work in custom loss functions. You can also still use a `idx` argument in custom loss functions.


gm89uk:
Hi @MilesCranmer,

Just FYI, I updated to master and my existing code runs significantly slower (about x10 longer days remaining with iterations=1,000,000) with the legacy code using idx in a custom loss function for batching=true. This is the only merged PR I can see that might have affected it.

MilesCranmer:
Many thanks for the report! I’ll take a look.

How does it change if you don’t have `idx` in your custom loss? This PR is built to make the batching built into the access of `.X`. But for compatibility it also allows the old explicit `idx` too.

gm89uk:
```julia
function loss_fnc(tree, dataset::Dataset{T,L}, options) where {T,L}    
    X = dataset.X
    y = dataset.y
    prediction, complete = eval_tree_array(tree, X, options)
    !complete && return L(Inf)
    weights = dataset.weights
    #remaining custom loss func
```
```julia
model = SRRegressor(
   #... remaining parameters
    expression_spec = expression_spec, #expression_spec is unchanged. 
    loss_function_expression = loss_fnc,
    batching=true,
)
```
This is what I did to test it, it runs exceptionally slow (slower than when I specify idx, even on the merged #master)

Before I was doing: 

```julia
function loss_fnc(tree, dataset::Dataset{T,L}, options, idx) where {T,L}    
    X = idx === nothing ? dataset.X : view(dataset.X, :, idx)
    prediction, complete = eval_tree_array(tree, X, options)
    !complete && return L(Inf)
    weights = idx === nothing ? dataset.weights : view(dataset.weights, idx)
```

so:
1.7.2, ETA is 9 days
#master with idx ~400 days
#master without specifying idx as per above code ~600 days.

MilesCranmer:
Just confirmed a 2x slowdown from this benchmark test:

```julia
SUITE["hack"] = @benchmarkable(
    equation_search(X, y; options, parallelism=:serial, niterations=100),
    setup=(
        options = Options(;
            binary_operators=(+, -, *, /),
            unary_operators=(exp, abs),
            maxsize=30,
            verbosity=0,
            progress=false,
            seed=0,
            deterministic=true,
            batching=true,
            batch_size=32,
        );
        rng = Random.MersenneTwister(0);
        X = randn(rng, 1, 1000);
        y = randn(rng, 1000);
    )
)
```

Will post in issue.

PR #426, opened 2025-02-22
fix: batched dataset for optimisation

### LINKED ISSUE #425: Performance regression for batching
As pointed out in #421 by @gm89uk, there's a big slowdown in performance for batching. Confirmed with the following benchmark:

```julia
SUITE["hack"] = @benchmarkable(
    equation_search(X, y; options, parallelism=:serial, niterations=100),
    setup=(
        options = Options(;
            binary_operators=(+, -, *, /),
            unary_operators=(exp, abs),
            maxsize=30,
            verbosity=0,
            progress=false,
            seed=0,
            deterministic=true,
            batching=true,
            batch_size=32,
        );
        rng = Random.MersenneTwister(0);
        X = randn(rng, 1, 1000);
        y = randn(rng, 1000);
    )
)
```

I added this to the benchmark/benchmarks.jl script and then ran AirspeedVelocity.jl with

```bash
 benchpkg -s benchmark/benchmarks.jl -f hack -r v1.7.2,dirty -a Bumper,LoopVectorization
```

The output is

```
|      | v1.7.2 | dirty  | v1.7.2/dirty |
|:-----|:------:|:------:|:------------:|
| hack | 11.8 s | 20.8 s | 0.568        |
```

so indeed a 2x slowdown. Looking into this now.

MilesCranmer:
Fixes #425

Basically the batching refactor resulted in doing full-batch optimisation when it previously did mini-batch.

cc @gm89uk

### DIFF
--- a/src/SingleIteration.jl
+++ b/src/SingleIteration.jl
@@ -74,6 +74,8 @@ function optimize_and_simplify_population(
     # to manually allocate a new task with a larger stack for Enzyme.
     should_thread = !(options.deterministic) && !(isa(options.autodiff_backend, AutoEnzyme))
 
+    batched_dataset = options.batching ? batch(dataset, options.batch_size) : dataset
+
     @threads_if should_thread for j in 1:(pop.n)
         if options.should_simplify
             tree = pop.members[j].tree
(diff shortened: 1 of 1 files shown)


MilesCranmer:
Want to give this a try?

gm89uk:
Thanks Miles,
It works great and actually there is a speed bump (presumably from not having to batch every eval). 

Thank you for such a quick fix.

MilesCranmer:
Awesome!

I'm hoping this also improves search performance too. Before that change, the per-population hall of fame would just end up being which expression got the "easiest batch" rather than the best expression. So now it fixes the batch for an entire iteration which makes the comparison more fair.

(The global hall of fame is still measured with the full batch, as before)

MilesCranmer:
Amazing!!

PR #440, opened 2025-04-21
[DRAFT] Abstract Class for PopMember

atharvas:
This adds an abstract type `AbstractPopMember`. Right now, there is no clean way to store additional metadata that is relevant for calculating statistics of how each population is evolving. This PR serves as a first step towards a provenance framework. It enables subtyping the pop-member object to track additional information about each population member node. 

Some issues (this PR should NOT be merged until we have concrete answers for these questions)
 - `AbstractPopMember`: Is this functionality desired in SR.jl?
 - The current code does not yield concrete types for many SR.jl functions (`get_pareto_frontier`, `extract_from_worker`, etc.). I've marked them as unstable right now. I'm hoping they don't have a big hit in benchmarking but if they do, then we'll need to take a closer look. 
 - Alternatively, we can use the `metadata` entry in `SymbolicExpression` objects instead. This is slightly unwieldy as I'll need to unpack and repack the PopMember objects during the `mutate!` calls when the metadata is updated. I want to consider this before moving forward.


MilesCranmer:
> * `AbstractPopMember`: Is this functionality desired in SR.jl?

Sounds good to me!

> * The current code does not yield concrete types for many SR.jl functions (`get_pareto_frontier`, `extract_from_worker`, etc.). I've marked them as unstable right now. I'm hoping they don't have a big hit in benchmarking but if they do, then we'll need to take a closer look.

Hm. There's probably a way to fix this. `@unstable` should only be used as a last resort.

> * Alternatively, we can use the `metadata` entry in `SymbolicExpression` objects instead. This is slightly unwieldy as I'll need to unpack and repack the PopMember objects during the `mutate!` calls when the metadata is updated. I want to consider this before moving forward.

I think we should just pick whatever is most semantically correct, and make it work. For `AbstractExpression` objects, they tend to only hold information that is independent of the dataset and search. So this is why the loss and cost are stored in the `PopMember` - because they are properties of the dataset and search, and should be updated if you pass a new dataset in, or restart the search. But expression objects are kind of independent of that sorta thing.

So with this in mind, I guess `PopMember` makes more sense for the types of metadata you'd want to attach?

src/SymbolicRegression.jl
```diff
@@ -560,10 +561,10 @@ end
     datasets::Vector{D}, ropt::AbstractRuntimeOptions, options::AbstractOptions, saved_state
 ) where {D<:Dataset}
     _validate_options(datasets, ropt, options)
-    state = _create_workers(datasets, ropt, options)
+    state = _create_workers(PopMember, datasets, ropt, options)
```
MilesCranmer, line 564: Ah. This might be why its unstable. This should probably be the fully parametrized type. Like `PopMember{T,L,N}` or whatever. 

I would actually recommend storing the pop member type in `options`, then you can just get it directly from there without needing to pass it around. So you would write this as `options.popmember_type{T,L,E}` or whatever.

Though ideally you would have a dedicated function like `with_type_parameters{::Type{<:AbstractPopMember}, options, dataset, ropt)` so that if someone's type has additional type parameters, they can override the behaviour.

atharvas: I see. This is similar to how `node_type` and `expresison_type` are passed around. I can rework the code with this in mind.

atharvas:
Sounds good. I'll make the necessary changes and push again.

MilesCranmer:
Ping on this; want any help just let me know

atharvas:
Hey sorry for the delay on this. I'm going to restart work in this and hope to have it in a merge-able state end of the week.

atharvas:
@MilesCranmer . Apologies but I don't think I'll have the bandwidth to work on this until the end of September. I'll close this PR then (hopefully in time for v2.1, if I can't make v2.0).

PR #441, opened 2025-04-25
Unused `TemplateExpression` in template_expression_complex.jl

Moelf:
```julia
ex = TemplateExpression(
    (; B_x, B_y, B_z, F_d_scale);
    structure, operators, variable_names
)
```

is un-used, should it be used?


MilesCranmer:
Thanks. Actually the `ex` is just supposed to be printed (it gets shown via Literate.jl). It is a different type from an expression spec though.

Moelf:
ah right, so maybe the doc should say "here's what will get constructed automatically when doing
```julia
TemplateExpressionSpec(; structure)
```

PR #442, opened 2025-04-25
fix `@template_spec` doc string

Moelf:
the original doc string results in parse error:
```julia
ParseError:
# Error @ ]8;;file:///home/jiling/.julia/environments/BjetTLA/notebooks/In[26]#6:1\In[26]:6:1]8;;\
    return p1[class] * g(x1^2) + f(x1, x2, p2[class]) - p3[1],
end
└─┘ ── invalid identifier
```


MilesCranmer:
Thanks!

MilesCranmer:
In retrospect I guess the example should really be an example... not put where the signature is. But can fix that later.

MilesCranmer (approved):

PR #447, opened 2025-05-06
feat: allow negative losses

MilesCranmer:
Requested a few times by @Moelf; finally added it.

This changes the scoring equation so needs to be accessed with the `loss_scaling` parameter

TODO:
- [x] Rename parameter to `loss_scaling`
- [x] Add a unittest

### DIFF
--- a/src/HallOfFame.jl
+++ b/src/HallOfFame.jl
@@ -219,34 +219,42 @@ function format_hall_of_fame(hof::HallOfFame{T,L}, options) where {T,L}
         end
     end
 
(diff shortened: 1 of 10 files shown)


MilesCranmer:
@Moelf the downside of this is that the "score" seems to not have much use anymore. Wondering if I should just not show that column at all... What do you think?

Moelf:
yeah the score isn't useful in this case.

we were using some kind of F-test (which is a function of N parameters and the chi2 loss) for comparing between different formulas on the Pareto front

PR #453, opened 2025-05-17
feat: allow `@template_spec` to declare explicit num_features

MilesCranmer:


gm89uk:
So does this allow a max 'complexity ' per expression? Otherwise, I'm not sure what this does as the number of features is defined by the template structure already?

If you have f(x1,x2), but f max features as 5, what does that do? Or conversely if you use f(x1,...,x5) and max features is 2, then it performs a sort of feature selection?

Thank you

MilesCranmer:
Oh this is just for cases where the automatic inference fails to figure out the number of features used by each expression. 99% of the time you won't need to use the `num_features` keyword

PR #468, opened 2025-06-26
full Mooncake integration

MilesCranmer:
@adil-soubki

Demo:

```julia
using SymbolicRegression, Mooncake, MLJBase, DifferentiationInterface

X = (; x=randn(100), y=randn(100), z=randn(100), w=randn(100))
y = @. 2 * cos(X.x) + X.y^2 - 4 * X.z + 3 * X.w

expression_spec = @template_spec(
    expressions = (f, g),
    parameters = (p=1,)
) do x, y, z, w
    return f(x, y) + g(z) + p[1] * w
end

model = SRRegressor(;
    binary_operators=(+, *, /, -),
    unary_operators=(cos, exp),
    autodiff_backend=AutoMooncake(; config=nothing),
    expression_spec=expression_spec,
)
mach = machine(model, X, y)
fit!(mach)
```

Output:

image

cc @yebai in case of curiosity.


gdalle:
Nice! Note that once we solve https://github.com/chalk-lab/Mooncake.jl/issues/472, most of the extension will hopefully become easier

MilesCranmer:
Appears this now needs https://github.com/chalk-lab/Mooncake.jl/pull/625.

(The MooncakeExt was removed from DispatchDoctor to avoid a circular dependency, but now Mooncake is erroring as it tries to differentiate through DispatchDoctor directives - which that PR should apply `@zero_adjoint` to)

yebai:
Happy to merge the MooncakeDD extension as a separate PR quickly.

MilesCranmer:
Thanks for offering! But I think it would result in too much extra work. For the time being I think we can just pin to earlier versions

src/ConstantOptimization.jl
```diff
@@ -93,38 +108,60 @@ function _optimize_constants(
...
+const CachedPrep = PerTaskCache{Dict{UInt,Any}}()
+
+@unstable function _cached_prep(ctx, backend, example_tree)
+    # We avoid hashing on the tree _value_ because it should not
+    # affect the prep. We want to cache as much as possible!
+    key = hash((ctx, backend, typeof(example_tree)))
+    stable_get!(CachedPrep[], key) do
+        prepare_gradient(ctx, backend, example_tree)
+    end
+end
```
MilesCranmer, line 156: @yebai @gdalle is there a robust way to cache the output of `prepare_gradient`? I was trying to figure out what sorts of things would invalidate said cache.

Basically I don't want to re-compute the prep for a new expression instance, since it's slow. So I cache the prep in `Base.task_local_storage()`, at a key based on (1) the dataset, (2) the backend, and (3) the expression's _type_.

But wasn't sure if this is going to break things or not.

Perhaps I can also cache based on the dataset's type, rather than the dataset's value?

MilesCranmer: Assuming this is ok

PR #487, opened 2025-08-12
slightly loosen ProgressMeter compat

oscardssmith:
We want to avoid 1.10.3, but I don't think there is a strong reason to prevent future patch releases.


MilesCranmer:
Unfortunately ProgressMeter.jl does not follow SemVer: https://github.com/timholy/ProgressMeter.jl/releases so I think we should treat every patch release as breaking. Especially since it's such a key dependency in both SymbolicRegression.jl and PySR, I really don't think we can afford such a risk

timholy:
@MarcMush (new features should use a minor version bump)

MarcMush:
@timholy yes I will be more careful

PR #505, opened 2025-09-07
Create `AbstractPopMember`

MilesCranmer:
@atharvas could you please review this as soon as you get a chance?

I did it from scratch and managed to get around the type instabilities. But I want to make sure it works for LaSR.

Look at `test/test_abstract_popmember.jl` for an example implementation. Basiclaly you need to implement `create_child` for one parent and two parents (crossover) as a way of propagating extra metadata forward. Or crossing over metadata in some way.


MilesCranmer:
Merged in other PR

PR #515, opened 2025-10-13
Adding optional argument to `create_child`

atharvas:
Includes the addition of a `mutation_choice` to `create_child` since in many cases the child creation is dependent on the type of mutation (e.g. if we are tracking the number of llm calls, we need to see if mutation_choice was an `llm_{choice}`)

One point that needs further discussion: Do we want to generalize this pattern? For my use case in LaSR all I need is the `mutation_choice` but we might want to generalize this. I'm not sure what would be the correct API to expose for this?

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -88,10 +88,10 @@ function create_utils_benchmark()
     suite["best_of_sample"] = @benchmarkable(
         best_of_sample(pop, rss, $options),
         setup = (
-            nfeatures = 1;
-            dataset = Dataset(randn(nfeatures, 32), randn(32));
-            pop = Population(dataset; npop=100, nlength=20, options=($options), nfeatures);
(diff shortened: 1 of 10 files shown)


MilesCranmer:
Thanks! I will think more about the right API.

@codex review

atharvas:
> The new _parse_guesses_impl now constructs members of an arbitrary P<:AbstractPopMember, but the very next step still calls strip_metadata(member, ...). strip_metadata only has methods for PopMember, so if the user supplies a custom popmember_type this call will hit a MethodError before guesses can be processed (same issue will surface in other call sites that strip metadata). Consider adding an overload of strip_metadata for AbstractPopMember (using create_child or similar) so custom population members can survive the warm‑start path.

looking into this... Seems like I'll need to make a new constructor for subtypes of AbstractPopMember

atharvas:
ok done. passes the test cases as well!

PR #524, opened 2025-11-07
Refactor test suite into projects

MilesCranmer:
This is inspired by Mooncake.jl's testing structure.

Also let's us more easily skip certain integration tests. For example, Mooncake.jl has been broken on 1.12 for a while (https://github.com/chalk-lab/Mooncake.jl/pull/714) and this is breaking the CI. With this refactoring we can simply skip Mooncake.


MilesCranmer:
@codex review

PR #562, opened 2026-02-12
fix: make stdin quit monitoring non-blocking

MilesCranmerBot:
Fixes intermittent hangs/freezes caused by stdin quit monitoring when stdin is a pipe/non-interactive (notably on Windows).

TDD: adds a fast unit reproducer using a FakeInputStream that simulates bytesavailable>0 with a blocking read, then fixes the monitor to use non-blocking reads + robust CRLF/partial-chunk parsing.

Related PySR issues: #832, #1101, #1092.


MilesCranmer (commented):
Looks actually pretty good.

However one thing I would like is for a test that does not rely on fake streams but on the actual REPL.

I think we can do this with Linux expect tool. Or are there other options?

@MilesCranmerBot please investigate and add such a test. 

I recommend doing some mutation testing as well - try introducing silent failures - does the test catch it? Or does the test always pass?

MilesCranmer:
@codex review

test/repl/test_repl_stdin_nonblocking.py
MilesCranmer: @MilesCranmerBot Can we do this with linux expect rather than hacking together a reimplementation

MilesCranmer:
@MilesCranmerBot please fix formatting

src/SearchUtils.jl
```diff
@@ -348,9 +348,46 @@ function init_dummy_pops(
     ]
 end
 
-struct StdinReader
+mutable struct StdinReader
     can_read_user_input::Bool
     stream::IO
```
MilesCranmer: @MilesCranmerBot should these two be marked `const` since you are changing to mutable?

test/unit/templates/test_template_macro.jl
```diff
@@ -3,8 +3,9 @@
     using DynamicExpressions: OperatorEnum, Node
 
     # Test basic parameter/expression handling
-    expr_spec = @template_spec(
-    parameters = (p1=10, p2=10, p3=1),  expressions = (f, g))  do x1, x2, class
+    expr_spec = @template_spec(parameters=(p1=10, p2=10, p3=1),  expressions=(f, g))  do x1,
+    x2,
+    class
```
MilesCranmer: @MilesCranmerBot this seems unrelated to the PR, if this is an issue maybe make a new PR for it? This formatting here looks a bit odd though, might be better to simply disable the formatter for this block of code

test/unit/templates/test_template_macro.jl
```diff
@@ -3,11 +3,13 @@
...
-    expr_spec = @template_spec(parameters=(p1=10, p2=10, p3=1),  expressions=(f, g))  do x1,
-    x2,
-    class
+    #! format: off
+    expr_spec = @template_spec(
+        parameters = (p1=10, p2=10, p3=1), expressions = (f, g)
+    ) do x1, x2, class
         return p1[class] * x1^2 + f(x1, x2, p2[class]) - g(p3[1] * x1)
     end
+    #! format: on
```
MilesCranmer: @MilesCranmerBot P.S., in parallel. Do you think this is a bug in JuliaFormatter.jl? Maybe worth checking different versions and pinning down what version this behavior changed. And then reporting it to the JuliaFormatter repo as an issue. Could you please do this.

Cuz I think
```julia
expr_spec = @template_spec(parameters=(p1=10, p2=10, p3=1),  expressions=(f, g))  do x1,
x2,
class

    x1 * 2
end
```
Is weird.

MilesCranmerBot: Checked this against JuliaFormatter releases and filed upstream: https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues/1018

Result: yes, this looks like a JuliaFormatter bug/regression around macro-call `do` blocks with keyword-style macro args. With `style=BlueStyle()`, v1.0.0 through v2.4.0 format the snippet, though awkwardly by splitting `expr_spec =` onto its own line. The behavior changes at v2.5.0: v2.5.0, v2.5.1, v2.5.2, and current `main` leave the macro call partially unformatted, preserving the double spaces and the bad multiline `do` argument layout. No SymbolicRegression.jl code changes needed here.

MilesCranmerBot: Checked this. Yes, I think the SciMLStyle output is bug-worthy.

Findings:
- The exact weirdness (`do x1,` followed by `x2,`/`class` at column 1) reproduces on JuliaFormatter v2.10.1 with `style=SciMLStyle(), margin=92`.
- It is *not* new in v2.10: I reproduced the unindented `do` args back to v1.0.55.
- There was a noticeable formatting behavior change between v2.4.0 and v2.5.0 for this macro-call case; v2.5/v2.6 also preserved the original double spaces and briefly emitted `x1*2` in the body. Latest v2.10.1 fixed the body spacing again, but the unindented `do` args remain.
- With SymbolicRegression's current `style = "blue"`, the same snippet formats cleanly, so this looks specific to SciMLStyle rather than the SR formatter config.

Reported upstream: https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues/1200

MilesCranmer:
@codex review

.github/workflows/CI.yml
```diff
@@ -90,6 +90,20 @@ jobs:
           julia --color=yes --project=. --threads=auto --check-bounds=yes --depwarn=yes --code-coverage=user -e 'import Coverage; import Pkg; Pkg.test(coverage=true)'
           julia --color=yes coverage.jl
         shell: bash
+      - name: "Install expect"
```
MilesCranmer: This should really be its own job... I'm not sure it needs to be part of the regular CI... @MilesCranmerBot

MilesCranmerBot: Done in 07cb7244: the REPL stdin monitoring check is now its own Ubuntu/Julia 1.10 CI job, rather than a step inside the `evaluation` unit-test matrix job. It stays out of coverage aggregation. Verified locally with the real `expect` test, including its intentional blocking negative control.

MilesCranmer:
@MilesCranmerBot can we actually reproduce the bug with this fix reverted? Like, are we sure this fixes the underlying issue? Just curious.

MilesCranmer (approved):

Issue #362, opened 2024-10-26
[BUG]: TaskFailedException when using eval_grad_tree_array and batching=true


gm89uk:
### What happened?

```julia

function loss_fnc(tree, dataset::Dataset{T,L}, options, idx) where {T,L}    
    # Extract data for the given indices
    X = idx === nothing ? dataset.X : view(dataset.X, :, idx)
    y = idx === nothing ? dataset.y : view(dataset.y, idx)
    weights = idx === nothing ? dataset.weights : view(dataset.weights, idx)
    prediction, grad, complete = eval_grad_tree_array(tree, X, options;variable=true)
    if !complete
        return L(Inf)
    end
    println("Size of grad: ", size(grad))
   #....remainingcode
```
This throws a nested task error: TaskFailedException when batching=true. Furthermore, size of grad: (6, 4927), rather than grad: (6, 50) for 6 features. 
The code works if I turn off batching.

If I switch to diff instead of grad for one feature:

```julia
  prediction, diff, complete = eval_diff_tree_array(tree, X, options,3)
    if !complete
        return L(Inf)
    end
    println("Size of diff: ", size(diff))
```

This now correctly works with batching and outputs the expected size of 50. 

The error triggers shortly after Info: Started!
I am on Julia version 1.10.4

Initialisation:
```julia
model = SRRegressor(
    niterations=1000000,
    binary_operators=[+,-,*,/],
    maxsize=60,
    bumper=true,
    turbo=true,
    #warm_start=true,
    populations=18,
    population_size=100,
    parsimony = 0.01,
    batching=true,
    loss_function = loss_fnc,
    #variable_names = ["a", "r", "p", "d", "l", "g"]
)
mach = machine(model, x_normalized, y,weightsIn)
fit!(mach)
```

I apologise if I am overlooking something.

### Version

v1.0.0-beta1

### Operating System

Windows

### Interface

Julia REPL

### Relevant log output

```shell
1-element ExceptionStack:
TaskFailedException
Stacktrace:
  [1] wait
    @ .\task.jl:352 [inlined]
  [2] fetch
    @ .\task.jl:372 [inlined]
[...]

PR #566, opened 2026-02-14
fix: eval_grad_tree_array works with SubArray inputs

MilesCranmerBot:
fixes a crash when calling `eval_grad_tree_array` with `X::SubArray` (e.g. `view(dataset.X, :, idx)` from batching or custom losses). on master this throws a `TypeError` because we typeassert the gradient to `typeof(X)`, while DynamicExpressions returns an `Adjoint{T, Matrix{T}}` for SubArray inputs.

the `Val(:eval_grad_tree_array)` method of `expected_array_type` now returns `typeof(similar(X))` instead of `typeof(X)`, and the gradient is normalized with `convert`. the convert is a no-op when the gradient already has the expected type (the `Matrix` path), and materializes the `Adjoint` only for view inputs. the return type stays concrete so DispatchDoctor stability checks are unaffected.

also adds a regression test exercising `eval_grad_tree_array` on `view(X, :, idx)`, compared against Zygote.

related: #362


src/InterfaceDynamicExpressions.jl
```diff
@@ -157,11 +156,11 @@ function DE.eval_grad_tree_array(
     kws...,
 )
     A = expected_array_type(X, typeof(tree))
-    dA = expected_array_type(X, typeof(tree), Val(:eval_grad_tree_array))
+    dA = Matrix{eltype(X)}
```
MilesCranmer: @MilesCranmerBot I'm not sure this is general enough though? Wouldn't want to force everything to be a Matrix. Perhaps a better solution is to have a cleaner `expected_array_type` that handles (in a generic way) things _like_ SubArray...

MilesCranmer (approved):

PR #602, opened 2026-05-13
Minimize plugin interface

adil-soubki:


adil-soubki:
This is superseded by #609.

PR #609, opened 2026-05-25
feat: composable plugin interface

MilesCranmer:
cc @adil-soubki

Changes the experimental plugin interface from #602 into a more composable shape: there is now a _tuple_ of plugins that are held in a heterogeneous tuple on `Options`, each carrying its own config and per-worker state.

Adaptive parsimony ships as the first proof plugin (`AdaptiveParsimonyPlugin`), and the old `RunningSearchStatistics` plumbing through `next_generation` / `best_of_sample` is gone.

Also, I removed the `@extend_mutation_weights` (will share cleaner alternative soon); `MutationWeights` users keep their kwargs but the engine no longer routes through `Symbol`-keyed dispatch. Backwards compat preserved for `mutation_weights`, `use_frequency`, `use_frequency_in_tournament`, etc.

Two follow-up PRs coming: one porting adaptive mutation weights to the same plugin shape, then one making mutations first-class composable objects (drops the `Val{:symbol}` dispatch entirely).


src/Mutate.jl
```diff
@@ -172,15 +199,24 @@ function condition_mutate_constant!(
 end
 
 # Go through one simulated options.annealing mutation cycle
+@inline function _fire_on_mutation_end!(
+    options::AbstractOptions, plugin_states::Tuple, event::MutationEvent, dataset
+)
+    for (plugin, pstate) in zip(options.plugins, plugin_states)
+        on_mutation_end!(pstate, plugin, event, dataset, options)
+    end
```
MilesCranmer, line 205: @adil-soubki these loops are so that one could set multiple plugins, via `Options(plugins=(MyPlugin1(), MyPlugin2()))`, and they would each be able to interact with the state of the search.

I think a lot of existing features in SR.jl can be rewritten as plugins and basically appended as "default plugins". It also helps ensure the plugin interface is generic. (This was done for adaptive parsimony)

src/Plugin.jl
```diff
@@ -0,0 +1,389 @@
...
+#                                   engine acts on. Composes by precedence:
+#                                   first non-default response in tuple order
+#                                   wins. Use for control-flow forks.
+#
+#   `init_X`                      — Factory (one-time). Called once per
+#                                   (plugin, worker) at startup. Returns a
+#                                   new instance.
+#
+#   `prepare_X`                   — Factory (per-context). Called per dispatch
+#                                   with context (output index, dataset).
```
MilesCranmer: This is what I was thinking for naming scheme, let me know what you think @adil-soubki!

src/plugins/AdaptiveParsimony.jl
```diff
@@ -0,0 +1,241 @@
...
+end
+
+function tournament_cost_multiplier(
+    s::AdaptiveParsimonyState,
+    p::AdaptiveParsimonyPlugin,
+    member::AbstractPopMember{T,L,N},
+    options::AbstractOptions,
+) where {T,L,N}
+    p.tournament || return one(L)
+    rss = s.rss[1]
```
MilesCranmer: Argh I just realised this is a bug. Fixing now... 

We need to have the plugin states be a vector (length == `nout`) of tuples (each tuple length == num plugins). This is because the `nout` is basically just batching the search over multiple datasets at once, so the plugin states should be independent.

MilesCranmer:
P.S., @atharvas, this will make it way easier to integrate LaSR by the way. Then someone can turn it on by literally doing `Options(plugins=(LaSR(),))`. And ideally it would be compatible with other plugins too.

src/plugins/AdaptiveParsimony.jl
```diff
@@ -0,0 +1,226 @@
...
+end
+
+"""
+    AdaptiveParsimonyState <: AbstractPluginState
+
+Per-output plugin state for [`AdaptiveParsimonyPlugin`](@ref). One instance
+per output (per dataset in multi-target regression); each instance owns a
+single `RunningSearchStatistics`.
+"""
+struct AdaptiveParsimonyState <: AbstractPluginState
```
MilesCranmer: @ [myself] we probably don't need an `AbstractPluginState` since dispatch is already performed using the `AbstractPlugin`, so that would simplify states so people could simply pass around named tuples or whatever

MilesCranmer:
@ [myself] I think the user should be allowed to pass a vector of plugins, and these would simply be turned into a tuple.

Also. We should have a `default_plugins` that can be overridable, similar to having a `default_mutations` in #610

PR #610, opened 2026-05-26
feat!: first-class AbstractMutation type hierarchy

MilesCranmer:
cc @adil-soubki @atharvas

Second piece of the plugin/mutation refactor. Mutations are now first-class types subtyping `AbstractMutation` (`MutateConstant`, `Backsolve`, `AddNode`, ..., 15 in total). `mutate!` dispatches on the concrete type instead of `Val{:symbol}` lookups, and the active mutation list is a `Vector{Pair{AbstractMutation, Float64}}` settable via `Options(; mutations = ...)`.

Mutation-specific config now lives on the mutation itself: `MutateConstant(; perturbation_factor, probability_negate)`, `Backsolve(; max_library_size, lambda, max_iter)`. The legacy `mutation_weights`, `perturbation_factor`, and `probability_negate_constant` kwargs still work via auto-conversion. `BacksolveOptions` was an internal alias and is gone.

`next_generation` calls into a per-mutation-type function barrier so the dynamic dispatch fires once per cycle. Benchmarks are within noise of the `Val{}`-dispatched version across 5 runs.

One follow-up PR coming: port adaptive mutation weights to the new API.


MilesCranmer:
Notes from meeting with @adil-soubki - we should have some kind of `default_mutations` option in `Options()` so that someone could easily turn off all the default mutations by simply passing `default_mutations=()`

PR #612, opened 2026-05-26
feat!: adaptive-mutation / annealing / mutation-loop plugins + per-cycle hooks

MilesCranmer:
cc @adil-soubki @atharvas

Third piece of the plugin/mutation refactor. Four new engine hooks (`on_cycle_start!`, `prepare_mutation_context`, `condition_mutation!`, `wrap_mutation_step`) and three plugins built on them:

- **`AdaptiveMutationWeightsPlugin`** is the one carrying empirical weight: per-mutation weights adapt online from accepted-and-improving outcomes, per-worker stats with no cross-population aggregation. Credit to Adil for the original idea and implementation; this PR promotes it to a full built-in plugin since it gave the largest search-quality lift we measured. `skip_in_adaptive_weights(::AbstractMutation)` controls which mutations sit out.

- **`MutationLoopPlugin`** layers two opt-in extensions on top: retrying rejected mutations up to a budget, and chaining additional mutations after a successful one. Retry outer, compound inner, one `wrap_mutation_step` impl. Empirically stacks on top of adaptive weights for a further boost.

- **`SimulatedAnnealingPlugin`** exists mostly to evacuate temperature from the engine. The plugin owns it on mutable state, scales `MutateConstant`'s perturbation through `condition_mutation!` on `MutateConstantContext`, and contributes `exp(-delta/(T*alpha))` via `mutation_acceptance_multiplier`. Auto-injected from `Options(; annealing, alpha)`. Bit-identical to the previous inline behavior. `annealing` and `alpha` fields move off `Options` and onto the plugin.

All three are opt-in. Defaults are tuned on one benchmark, flagged extra-experimental in docstrings.


MilesCranmer:
@MilesCranmerBot can you please rewrite this PR on top of the latest master? Then please submit it as a new PR.

PR #544, opened 2026-02-04
ci: allow-fail enzyme integration group

MilesCranmerBot:
Allow the Enzyme AD integration group to fail without failing CI (continue-on-error for matrix.group == "ad/enzyme").\n\nThis keeps the job running and reporting, but unblocks master when Enzyme flakes/breaks.

PR #613, opened 2026-05-27
docs: wording tweaks for ad backends

MilesCranmer:
c.f. #544

FYI @wsmoses

### DIFF
--- a/src/Options.jl
+++ b/src/Options.jl
@@ -438,8 +438,9 @@ const OPTION_DESCRIPTIONS = """- `defaults`: What set of defaults to use for `Op
     an instance of `AbstractADType` (see `ADTypes.jl`).
     Default is `nothing`, which means `Optim.jl` will estimate gradients (likely
     with finite differences). You can also pass a symbolic version of the backend
-    type, such as `:Enzyme` for Enzyme.jl (*experimental/unstable*; may fail on
-    some platforms and may be allowed to fail in CI), `:Mooncake` for Mooncake.jl,
+    type, such as `:Enzyme` for Enzyme.jl (SR.jl's Enzyme bridge is
+    *experimental*; may fail on some platforms and may be allowed to fail in
+    CI), `:Mooncake` for Mooncake.jl,
     or `:Zygote` for Zygote.jl. Most backends will not work, and many will never
     work due to incompatibilities, though support for some is gradually being added.
 - `perturbation_factor`: When mutating a constant, either
--- a/src/Options.jl
+++ b/src/Options.jl
@@ -438,10 +438,9 @@ const OPTION_DESCRIPTIONS = """- `defaults`: What set of defaults to use for `Op
     an instance of `AbstractADType` (see `ADTypes.jl`).
     Default is `nothing`, which means `Optim.jl` will estimate gradients (likely
     with finite differences). You can also pass a symbolic version of the backend
-    type, such as `:Enzyme` for Enzyme.jl (SR.jl's Enzyme bridge is
-    *experimental*; may fail on some platforms and may be allowed to fail in
-    CI), `:Mooncake` for Mooncake.jl,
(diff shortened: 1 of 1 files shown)


MilesCranmer:
@wsmoses let me know if this is good?

MilesCranmer:
@wsmoses 

Also. Would it help if I get my agent to monitor the Enzyme integration tests and send over a MWE when there are failures? Or would it be better if the builtin integration tests were expanded to include the other pieces (like parametric expressions - which is where this issue seems to be coming from).

wsmoses:
so even better, I've fixed the mwe from earlier. Can we check if it now works [and if so just have it listed normally]

MilesCranmer:
That's great, thanks!!

Re: docs, the plugin is still experimental so I wouldn't want to fully remove the warning yet, until there is feature parity. For example, `parallelism=:multiprocessing` is still incompatible at the moment.

wsmoses:
hm got any current mwes for that offhand?

MilesCranmer:
On Julia 1.12.6:

```julia
using SymbolicRegression, Enzyme

options = Options(; autodiff_backend=:Enzyme)

equation_search(
    randn(1, 1),
    randn(1);
    options,
    niterations=1,
    parallelism=:multiprocessing,
    numprocs=1,
)
```

with error:

```julia
From worker 2: VT: i64 idxs:{} start=0 size=4 storeSize=8 ...
From worker 2: Assertion failed: (0 && "unhandled accumulate with partial sizes"), function addToDiffe, file .../DiffeGradientUtils.cpp, line 433.
```

wsmoses:
Can you send the llm on a mission to make an mwe that ideally doesn't use external packages and file an issue

MilesCranmer:
It's odd because this is identical code to the branch that works (`parallelism=:multithreading`), except for the fact that it is running inside a Distributed.jl process. 

My LLM agent seems to be unable to create a believable MWE of this, unfortunately. I can still make an issue with this error if you want though.

wsmoses:
go ahead and make an issue. In the meantime, we can change the wording like this for enzyme and mooncake [which isn't run at all for 1.12, and doesn't support multithreading either]

MilesCranmer:
Just to be clear, the multiprocessing is in the outer loop; there is no multithreading/multiprocessing inside the function being differentiated.

MilesCranmer:
> On Julia 1.12.6:
> 
> ```julia
> using SymbolicRegression, Enzyme
> 
> options = Options(; autodiff_backend=:Enzyme)
> 
> equation_search(
>     randn(1, 1),
>     randn(1);
>     options,
>     niterations=1,
>     parallelism=:multiprocessing,
>     numprocs=1,
> )
> ```
> 
> with error:
> 
> ```julia
> From worker 2: VT: i64 idxs:{} start=0 size=4 storeSize=8 ...
> From worker 2: Assertion failed: (0 && "unhandled accumulate with partial sizes"), function addToDiffe, file .../DiffeGradientUtils.cpp, line 433.
> ```

@MilesCranmerBot can you please locally reproduce this error and then create an appropriate issue in Enzyme.jl

MilesCranmer:
And good point about Mooncake, I will mark that experimental too until 1.12 is working.

wsmoses:
@MilesCranmer so this should now be fixed enzyme side, can you retest?

wsmoses:
bump here!

MilesCranmer:
Bot ran out of tokens it seems; I'll look later next week

wsmoses:
bump @MilesCranmer !

MilesCranmer:
Just tried; so far so good, thanks. Will set up the CI again and let you know if there are any new issues popping over the next couple of months

MilesCranmer:
@MilesCranmerBot could you set up the i/ad/enzyme test for both Julia 1.10 and 1.12 and turn off allow-fail? (Make a PR for this)

PR #614, opened 2026-05-30
ci: pin JuliaFormatter to 2.4.0 in format workflows

MilesCranmer:
`check-format` pinned JuliaFormatter to a floating `@2`, so CI used whatever 2.x patch was newest at run time. JuliaFormatter 2.5.x changed formatting (e.g. it rewraps multi-line signatures), so files across the repo now fail the check despite being unchanged from master, e.g. `src/CheckConstraints.jl`. master itself would fail `check-format` if re-triggered.

this pins `check-format` and `fix-format` to `2.4.0`, the version the repo is currently formatted against (verified `format(".")` is a no-op on master). `fix-format` was separately stuck on `1.0.61`, which formats to v1 style and would fight the v2 checker.

### DIFF
--- a/.github/workflows/check-format.yml
+++ b/.github/workflows/check-format.yml
@@ -25,7 +25,7 @@ jobs:
         uses: julia-actions/cache@v3
       - name: Install JuliaFormatter and format
         run: |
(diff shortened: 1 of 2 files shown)


MilesCranmer:
We can revert this after https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues/1013

However I think it'd be easier to just leave the formatter version number pinned to not generate more maintenance burden.

PR #618, opened 2026-06-16
docs: update equation_search docstring

spinnau:
The `equation_search` docstring was missing description for the `run_id` and `extra` arguments. Thus, I have copied the descriptions for these arguments from `MLJInterface.jl` and `Dataset.jl`.


MilesCranmer:
Good catch! Thanks

@codex review

PR #619, opened 2026-06-18
Add TemplateExpression shared-state optimization hooks

adil-soubki:
Introduce a template-level plugin interface for candidate-local state that is shared across the inner expressions of a TemplateExpression. This supports plugins whose optimizable state belongs to the whole template candidate rather than to each inner expression independently.

The new shared-state hooks are:

- NoTemplateSharedState
- template_shared_state
- get_template_shared_state
- copy_template_shared_state
- attach_template_shared_state

TemplateExpression construction now creates one shared state object for the candidate and attaches it to each inner expression. The same invariant is preserved when templates are copied, copied into preallocated storage, reconstructed with with_contents, mutated through with_contents_for_mutation, or rebuilt through crossover. The default hook behavior is a no-op, preserving existing TemplateExpression behavior for normal ComposableExpression inners.

Add template-level constant optimization hooks:

- get_template_constants_for_optimization
- set_template_constants_for_optimization!
- extract_template_gradient_for_optimization

The default implementations reproduce the previous behavior by concatenating inner-expression optimization constants with template parameters. Plugins can override these hooks when candidate-local template state should appear only once in the optimizable constant vector. count_constants_for_optimization now derives from get_constants_for_optimization so custom hook-provided constants stay in sync with optimizer buffer sizing.

Update optimization-gradient extraction for TemplateExpression and ComposableExpression so wrapped expressions and structural AD tangents are handled by the optimization API. The template fallback now accepts gradients with trees/metadata fields as well as DynamicExpressions accessors, and the ComposableExpression fallback unwraps the tree gradient before delegating to the existing expression gradient machinery.

Preserve Mooncake-specific gradient behavior by adding extension methods for Mooncake tangents. ComposableExpression optimization gradients delegate back to the existing Mooncake DynamicExpressions extraction path, while TemplateExpression gradients extract inner and parameter tangents from Mooncake's fields layout. The existing Mooncake extract_gradient implementation now sizes its scalar constant buffer using count_scalar_constants rather than optimization-hook constant counts, keeping scalar-gradient extraction separate from custom template optimization constants.

Add regression coverage for:

- shared template state attachment during construction
- copy, copy_into!, with_contents, mutation reconstruction, and crossover
- parse_expression and create_expression with custom template inner types
- template-level get/set/extract optimization hooks
- optimization constant counts matching hook-provided constants
- ComposableExpression optimization-gradient extraction from wrapped trees
- TemplateExpression gradient extraction from accessor-style and structural tangents
- Mooncake optimization gradients for templates and custom inner wrappers

Tests run:

- julia --project=. -e 'using Pkg; Pkg.test(; test_args=["unit/evolution-core"])'
- julia --project=. -e 'using Pkg; Pkg.test(; test_args=["unit/expressions"])'
- julia --project=. -e 'using Pkg; Pkg.test(; test_args=["integration/ad/mooncake"])'
- git diff --check

### DIFF
--- a/ext/SymbolicRegressionMooncakeExt.jl
+++ b/ext/SymbolicRegressionMooncakeExt.jl
@@ -2,13 +2,46 @@ module SymbolicRegressionMooncakeExt
 
 using DynamicExpressions: DynamicExpressions as DE
 using SymbolicRegression: SymbolicRegression as SR
-using SymbolicRegression.ConstantOptimizationModule: count_constants_for_optimization
 using Mooncake: Mooncake
+
+function SR.ComposableExpressionModule._composable_gradient_tree(
+    gradient::Mooncake.Tangent
+)
+    return gradient.fields.tree
+end
+
+function SR.ComposableExpressionModule._extract_composable_gradient_for_optimization(
+    gradient::Mooncake.Tangent, ex::SR.ComposableExpression
+)
+    return DE.extract_gradient(gradient, ex)
+end
+
+function SR.TemplateExpressionModule._template_inner_gradient(
+    gradient::Mooncake.Tangent, ::SR.TemplateExpression, key::Symbol
+)
+    tree_gradient = getproperty(gradient.fields.trees, key)
+    @assert(
+        !(tree_gradient isa Mooncake.NoTangent),
+        "Unexpected input type: $(tree_gradient)::$(typeof(tree_gradient))"
+    )
+    return tree_gradient
+end
+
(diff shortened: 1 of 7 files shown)


adil-soubki:
These format errors are on files that I didn't touch. It seems that CI is now using an updated version of JuliaFormatter (v2.8.5) and this changes output on old files?

adil-soubki:
Superseded by #644, which replaces the shared mutable template-state
approach with template-owned generic optimizable parameters.

PR #633, opened 2026-07-14
Show real variable names in template expression printing

### LINKED ISSUE #456: Template expressions printing should show variables
I think template expressions should show variables explicitly, provided an expression object is not used multiple times, or used on raw numerical input.

singhharsh1708:
## Summary
`TemplateExpression`'s `string_tree` always printed features as generic `#1`, `#2`, ... placeholders, even though the real dataset variable names are already stored in the expression's metadata (`get_metadata(ex).variable_names`).

This uses the real names when their count matches the total number of features referenced by the template's sub-expressions, and falls back to the generic `#i` labels otherwise (e.g. when variable names aren't available).

Fixes #456

## Test plan
- [x] Existing `test/unit/templates/test_template_expression_string.jl` passes (7/7)
- [x] Verified manually: a template expression built with `variable_names=["x"]` now prints `f = x` instead of `f = #1`

### DIFF
--- a/src/TemplateExpression.jl
+++ b/src/TemplateExpression.jl
@@ -603,7 +603,13 @@ function DE.string_tree(
     expressions = get_contents(ex)
     num_features = get_metadata(ex).structure.num_features
     total_num_features = max(values(num_features)...)
-    variable_names = ['#' * string(i) for i in 1:total_num_features]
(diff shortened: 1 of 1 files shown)


MilesCranmer:
This is incorrect; someone could still use `f(1.0)` for example. I don't believe this is fixable in a safe and general way.

PR #634, opened 2026-07-14
Allow real-valued weights with complex-valued X/y

### LINKED ISSUE #477: [Feature]: Allow `weights` to have a different type than `y`
### Feature Request

Hi!

**My use case**: Complex-valued `y` with scalar weights.

There's a simple **workaround**: `weights = weights .+ 0im`, but I was wondering why this is not the default mode of operation.

Thanks!

singhharsh1708:
## Summary
`equation_search`'s `weights` keyword was type-constrained to match the element type of `X`/`y`, so passing real-valued weights (e.g. `Float64`) alongside complex-valued data (e.g. `ComplexF64`) failed with a `TypeError` at the top level, and would then hit an explicit error inside `_weighted_loss` even after loosening that constraint:

```
Element type of `x` is ComplexF64, element type of `y` is ComplexF64, and element type of `w` is Float64. All element types must be the same.
```

There's no numerical reason to require this — a real weight scaling a complex residual/loss is well defined. This:
- loosens the `weights` keyword signature in `equation_search` to accept any `AbstractVecOrMat`
- updates `_weighted_loss` to accept a weight array whose element type differs from `x`/`y`'s, as long as it's a subtype of `Real` (otherwise it still errors, now with a clearer message)

Fixes #477

## Test plan
- [x] Reproduced the original issue's example (complex `X`/`y` with real `weights` via `equation_search`) — now runs to completion instead of erroring
- [x] Added regression tests to `test/unit/misc/test_losses.jl`: real weights with complex `x`/`y` succeed; non-real, non-matching weight types still error
- [x] `test/unit/misc/test_losses.jl` passes (10/10)

### DIFF
--- a/src/LossFunctions.jl
+++ b/src/LossFunctions.jl
@@ -47,14 +47,8 @@ function _loss(
 end
 
 function _weighted_loss(
-    x::AbstractArray{T}, y::AbstractArray{T}, w::AbstractArray{W}, loss::LT
-) where {T,W,LT<:Union{Function,SupervisedLoss}}
-    if W !== T && !(W <: Real)
-        return error(
-            "Element type of `x` is $(T), element type of `y` is $(T), and element type of `w` is $(W). " *
-            "The weight type must either match `x`/`y`, or be a subtype of `Real`.",
-        )
-    end
+    x::AbstractArray{T}, y::AbstractArray{T}, w::AbstractArray, loss::LT
+) where {T,LT<:Union{Function,SupervisedLoss}}
     if loss isa SupervisedLoss
         return sum(loss, x, y, w; normalize=true)
     else
--- a/src/LossFunctions.jl
+++ b/src/LossFunctions.jl
@@ -47,7 +47,7 @@ function _loss(
 end
 
 function _weighted_loss(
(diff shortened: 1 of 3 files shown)


src/LossFunctions.jl
```diff
@@ -47,8 +47,14 @@ function _loss(
 end
 
 function _weighted_loss(
-    x::AbstractArray{T}, y::AbstractArray{T}, w::AbstractArray{T}, loss::LT
-) where {T,LT<:Union{Function,SupervisedLoss}}
+    x::AbstractArray{T}, y::AbstractArray{T}, w::AbstractArray{W}, loss::LT
+) where {T,W,LT<:Union{Function,SupervisedLoss}}
+    if W !== T && !(W <: Real)
```
MilesCranmer: This check looks a bit fragile

MilesCranmer: But maybe its required...?

singhharsh1708: Not required — _eval_loss casts to ::L at the call site anyway. Dropped the check.

MilesCranmer (changes requested):
I don't think this is actually working yet; needs integration testing.

I'm not sure this is the right fix tbh. Better solution would be to just promote types.

Issue #509, opened 2025-09-18
[BUG]:  [Julia] Could not regenerate all hof's equations when using guesses


qizzhao:
### What happened?

Hi, I have post the issue on #https://github.com/MilesCranmer/PySR/discussions/1032. 

SymbolicRegression package in Julia, which will generate equations and store as hall_of_fame.csv after training.  I am trying to reload all h.o.f equations and do predictions on new dataset X. Working on reloading the equations, it seems that reloaded model will ignores several h.o.f's equations after fitting. Below is a very short example: 
  - There is a hall_of_frame.csv with equations:  A, B, C, D, E
  - tried to use those equations above to do prediction for another dataset X. 
  - load hof.csv file and regenerate the Machine
  - new Machine only contains A, B, C and ignores D, E

Dataset are attached: 
 - `x.csv`:  dataset x for first time training
 - `y.csv`: dataset y for first time training
 - `hof.csv`: first time training output's equations

The julia scripts are in extra info below(You might need to copy and paste to file):  
 - `operator.jl`: binary and unary operators using during training
 - `loss_function.jl`: define loss function 
 - `script.jl`: main function to regenerate this situation. Including 2 parts:  `first_time_train` and `reload_model_and_predict_on_new_data`

You could regenerate the case through running `julia script.jl` and it will print out the number of hof's equation after fitted. 

If you need more information, please let me know  

[hall_of_fame.csv](https://github.com/user-attachments/files/22399332/hall_of_fame.csv)
[x.csv](https://github.com/user-attachments/files/22399331/x.csv)
[y.csv](https://github.com/user-attachments/files/22399333/y.csv)

### Version

v2.0.0-alpha.8

### Operating System

Linux

### Interface

Script (i.e., `python my_script.py`)

### Relevant log output

```shell

```

### Extra Info

> loss_function.jl

```
using  Statistics
using  SymbolicRegression


function mse_loss(tree, dataset::SymbolicRegression.Dataset{T,L}, options)::L where {T,L}

    function std(vec)
        return (sum( (vec.-sum(vec)/length(vec)).^2 ) / length(vec))^0.5
    end

    prediction, flag = eval_tree_array(tree, dataset.X, options)
    
    if (!flag) || (std(prediction)<1e-8)
        prediction = nothing
        return L(Inf)
    end
    
    loss =  sum((prediction .- dataset.y) .^ 2) / dataset.n
    
    prediction = nothing
    
    return loss

end
```

[...]

PR #636, opened 2026-07-14
Fix constant-optimization restarts never escaping a zero-valued constant

### LINKED ISSUE #510: [BUG]: Cannot load guesses in Template Expression if function contains parameters
### What happened?

For TemplateExpressions, I cannot load guesses that contain a parameter:
`guesses = [(f="(((#1 - #2) * -0.16528662) + #4) - #3")],`

In this instance #4 is p1[n]:
```julia
    expression_spec = @template_spec(expressions = (f,),parameters=(p1=n,)) do x1,x2,x3,x4 #x4 is param, #n is number of categories in x4
        LF = p1[x4]
	return f(x1,x2,x3,LF)
 end
```

The search runs no problem, but I cannot load the model again. Thank you in advance. 

### Version

v2.0.0-alpha.8

### Operating System

Windows

### Interface

Julia REPL

### Relevant log output

```shell
     
ERROR: syntax: misplaced assignment statement in "[f = "(((#1 - #2) * -0.16528662) + #4) - #3"]" around REPL[42]:1
Stacktrace:
 [1] top-level scope
   @ REPL[42]:1

julia> mach = machine(model, X, y)
ERROR: UndefVarError: `model` not defined in `Main`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
 [1] top-level scope
   @ REPL[43]:1
```

singhharsh1708:
## Summary
Root-caused the long-running #510/#509 thread (\"Cannot load guesses in Template Expression if function contains parameters\"). The community discussion converged on \"the optimizer seems to be failing somehow\" but never found the mechanism. Here it is:

`_optimize_constants_inner`'s restart loop perturbs the initial point **multiplicatively**:

```julia
xt = @. x0 * (ET(1) + ET(1 // 2) * eps)
```

If `x0[i] == 0`, then `xt[i] = 0 * (1 + 0.5*eps) = 0` for **every** restart, regardless of `eps`. So `optimizer_nrestarts` gives any zero-valued constant zero actual restarts — it's stuck exactly where it started.

This silently breaks guesses for `TemplateExpression`s with parameters specifically: a freshly-parsed guess always zero-initializes its parameters (`TemplateExpression.jl`'s constructor defaults `parameters` to `zeros(T, n)`), and `_parse_guesses_impl` gives it exactly **one** `optimize_constants` call before it enters the population. A normal population member gets many further chances via mutation across generations and can eventually escape a bad start; a guess does not. If that single shot can't move the zero-initialized parameters (e.g. because a `1/p1[cat]`-shaped term blows up to `Inf` right at `p1=0`, as in #509/#510's repros), the guess is permanently stuck.

Verified this exactly reproduces the reported failure mode and that the fix resolves it (repro modeled on #510's example: `y = p1[cat]*x^3 + sqrt(x)/p1[cat]`, `p1_true = [5.0, 10.0, 0.8, 3.1]`):

| | Loss | Recovered parameters |
|---|---|---|
| Before fix | `Inf` (stuck) | `[0.0, 0.0, 0.0, 0.0]` |
| After fix | `4.8e-21` | `[4.9999978, 10.0, 0.79999996, 3.09999996]` |

The fix switches the restart perturbation to also work from zero:

```julia
xt = @. x0 + ET(1 // 2) * eps * (abs(x0) + ET(1))
```

This is approximately the same relative-scale jitter as before for `|x0|` large, but adds an absolute floor so zero-valued constants actually get perturbed.

Fixes #510

## Test plan
- [x] Added a regression test in `test/unit/misc/test_guesses.jl` modeled directly on #510's repro (`TemplateStructure` with 4 categorical parameters, `1/p1[cat]` term) — fails with `loss == Inf` on master, passes with the fix
- [x] `test/unit/misc/test_guesses.jl` — 62/62 pass
- [x] `test/unit/evolution-core/test_custom_type_constant_optimization.jl` (constant-optimization restart path, custom types) — 2/2 pass

### DIFF
--- a/src/ConstantOptimization.jl
+++ b/src/ConstantOptimization.jl
@@ -90,7 +90,7 @@ function _optimize_constants_inner(
     for _ in 1:(options.optimizer_nrestarts)
         ET = eltype(x0)
         eps = randn(rng, ET, size(x0)...)
-        xt = @. x0 * (ET(1) + ET(1 // 2) * eps)
+        xt = @. x0 + ET(1 // 2) * eps * (abs(x0) + ET(1))
         tmpresult = Optim.optimize(obj, xt, algorithm, optimizer_options)
         num_evals += tmpresult.f_calls * eval_fraction
         # TODO: Does this need to take into account h_calls?
--- a/test/unit/misc/test_guesses.jl
+++ b/test/unit/misc/test_guesses.jl
@@ -107,6 +107,57 @@ end
     @test length(get_metadata(member.tree).parameters.p._data) == 2
 end
 
+@testitem "Guess with template parameters actually optimizes them" begin
+    # Regression test for #510: template parameters are always
+    # zero-initialized when parsing a guess, and get only a single shot at
+    # `optimize_constants`. The restart mechanism used to perturb constants
+    # multiplicatively (`x0 * (1 + eps)`), which can never move a
+    # zero-valued constant away from zero, so guesses depending on nonzero
+    # parameters were effectively unoptimizable.
+    using SymbolicRegression
+    using SymbolicRegression: calculate_pareto_frontier
+    using Random: MersenneTwister
+    using Test
+
+    structure = TemplateStructure{(:f,),(:p1,)}(
+        ((; f), (; p1), (x, cat)) -> f(x, p1[cat]); num_parameters=(; p1=4)
+    )
+    options = Options(;
+        binary_operators=[+, -, *, /],
+        unary_operators=[sqrt],
+        expression_spec=TemplateExpressionSpec(; structure),
+        optimizer_nrestarts=8,
+        deterministic=true,
+        seed=0,
+        verbosity=0,
+        progress=false,
+    )
+
+    rng = MersenneTwister(0)
+    n = 200
+    x_data = rand(rng, n) * 10.0
+    cat_data = rand(rng, 1:4, n)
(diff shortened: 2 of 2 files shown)


src/ConstantOptimization.jl
```diff
@@ -90,7 +90,7 @@ function _optimize_constants_inner(
     for _ in 1:(options.optimizer_nrestarts)
         ET = eltype(x0)
         eps = randn(rng, ET, size(x0)...)
-        xt = @. x0 * (ET(1) + ET(1 // 2) * eps)
+        xt = @. x0 + ET(1 // 2) * eps * (abs(x0) + ET(1))
```
MilesCranmer, line 93: This is changing the algorithm.

MilesCranmer:
I think it's solvable, but in order for it to not be a breaking change, it would need to special case `== 0` and e.g., switch to sampling N(0,1).

singhharsh1708:
Fair point — pushed a narrower fix that only special-cases `x0 == 0` (sampling from N(0,1) directly), leaving the existing multiplicative jitter untouched for everything else. Not an algorithm change for the nonzero case anymore. Still fixes the original repro (loss Inf -> ~2e-21, parameters recovered to 6 decimal places).

singhharsh1708:
Reopen was blocked, so opened a narrower follow-up addressing this feedback: #637 (only special-cases `x0 == 0`, no change for the nonzero restart path).

PR #637, opened 2026-07-15
Fix constant-optimization restarts never escaping a zero-valued constant (v2)

singhharsh1708:
Follow-up to #636 (closed), which Miles flagged as changing the restart algorithm for *all* constants, not just zero-valued ones. This version narrows the fix to only special-case \`x0 == 0\`:

\`\`\`julia
xt = @. ifelse(x0 == 0, eps, x0 * (ET(1) + ET(1 // 2) * eps))
\`\`\`

For any nonzero constant, behavior is byte-for-byte identical to before. Only the zero case changes: it now samples directly from \`N(0,1)\` (via the already-computed \`eps\`) instead of computing \`0 * (1 + 0.5*eps) == 0\` every restart.

## Summary (same root cause as #636)
\`_optimize_constants_inner\`'s restart loop perturbed the initial point multiplicatively: \`xt = x0 .* (1 + 0.5*eps)\`. If \`x0[i] == 0\`, this is exactly \`0\` for every restart regardless of \`eps\`, so \`optimizer_nrestarts\` gave zero-valued constants no actual restarts at all.

This silently broke guesses for \`TemplateExpression\`s with parameters (#510/#509): a freshly-parsed guess always zero-initializes its parameters, and \`_parse_guesses_impl\` gives it exactly one \`optimize_constants\` call before it enters the population — if that one shot can't escape a bad basin (e.g. a \`1/p1[cat]\` term blowing up near \`p1=0\`), the guess is permanently stuck.

Verified against #510's repro (\`y = p1[cat]*x^3 + sqrt(x)/p1[cat]\`, \`p1_true = [5.0, 10.0, 0.8, 3.1]\`):

| | Loss | Recovered parameters |
|---|---|---|
| Before fix | \`Inf\` (stuck) | \`[0.0, 0.0, 0.0, 0.0]\` |
| After fix | \`1.9e-21\` | \`[4.999999999999952, 9.999999999999849, 0.7999999999998596, 3.1000000000000996]\` |

Helps with #510

## Test plan
- [x] Regression test in \`test/unit/misc/test_guesses.jl\` modeled on #510's repro
- [x] \`test/unit/misc/test_guesses.jl\` — 60/60 pass
- [x] \`test/unit/evolution-core/test_custom_type_constant_optimization.jl\` (exercises the restart path with custom types) — 2/2 pass


MilesCranmer:
@codex review
@MilesCranmerBot review

src/ConstantOptimization.jl
```diff
@@ -90,7 +90,7 @@ function _optimize_constants_inner(
     for _ in 1:(options.optimizer_nrestarts)
         ET = eltype(x0)
         eps = randn(rng, ET, size(x0)...)
-        xt = @. x0 * (ET(1) + ET(1 // 2) * eps)
+        xt = @. ifelse(x0 == 0, eps, x0 * (ET(1) + ET(1 // 2) * eps))
```
MilesCranmer:
```suggestion
        xt = @. ifelse(iszero(x0), eps, x0 * (ET(1) + ET(1 // 2) * eps))
```

test/unit/misc/test_guesses.jl
```diff
@@ -107,6 +107,57 @@ end
...
+    hof = equation_search(
+        X,
+        y;
+        niterations=0,
+        options,
+        guesses=[guess],
+        variable_names=["x", "cat"],
+        parallelism=:serial,
+    )
+    dominating = calculate_pareto_frontier(hof)
```
MilesCranmer: This is the old API

test/unit/misc/test_guesses.jl
```diff
@@ -107,6 +107,57 @@ end
...
+
+    # Before the fix, the parameters stay at their zero-initialized values,
+    # the `1/p1[cat]` term blows up, and loss never leaves Inf.
```
MilesCranmer: I think relying on this for the test isn't a long-term solution unfortunately

test/unit/misc/test_guesses.jl
```diff
@@ -107,6 +107,57 @@ end
...
+    # zero-valued constant away from zero, so guesses depending on nonzero
+    # parameters were effectively unoptimizable.
+    using SymbolicRegression
+    using SymbolicRegression: calculate_pareto_frontier
+    using Random: MersenneTwister
+    using Test
+
+    structure = TemplateStructure{(:f,),(:p1,)}(
+        ((; f), (; p1), (x, cat)) -> f(x, p1[cat]); num_parameters=(; p1=4)
+    )
```
MilesCranmer: Please use new API for this

MilesCranmerBot (approved):
Reviewed `f8728b89cf` — no issues found.

I checked the constant-restart change and the regression test. The zero-only special case preserves the existing multiplicative jitter for nonzero constants while allowing zero-initialized template parameters to escape zero.

Verification:
- Local: `/usr/local/bin/julia --project=test test/runtests.jl unit/misc` → 114 passed
- PR checks: currently passing

test/unit/misc/test_guesses.jl
```diff
@@ -107,6 +107,62 @@ end
...
+    # Before the fix, the parameters stay at their zero-initialized values
+    # (the `1/p1[cat]` term blows up and loss never leaves Inf), so check
+    # directly that the recovered parameters are close to the true values,
+    # rather than relying on an opaque loss threshold.
+    best = argmin(m -> m.loss, dominating)
+    recovered_p1 = get_metadata(best.tree).parameters.p1._data
+    @test isfinite(best.loss)
+    @test maximum(abs.(recovered_p1 .- p1_true)) < 1e-3
+end
+
```
MilesCranmer:
```suggestion
```

MilesCranmer (approved):

Issue #363, opened 2024-10-27
Personalised compilation settings for running in Julia vs PySR


MilesCranmer:
It would be nice to have the precompilation settings vary by intended user - whether it be PySR users or Julia users.

For example, the Julia users probably want the MLJ interface precompiled.
And the PySR would only want the `equation_search` interface precompiled, since that's what PySR calls.

For a future issue: it would also be nice if we could precompile some of the Python->Julia calls in PythonCall.jl so speed up startup time.

ATOMSLab:
This looks like a much-needed fix for us! Thanks for looking into it!

In Issue #358 , we were never certain if the `saved_state` was importing correctly, because the method we were using to test this (setting most mutation weights to zero, so expression length would never change) doesn't work. We now think this is because precompiling is interfering with our wish to set most mutation weights to zero. 

This is also reflected in Issue #361 , where turning off precompilation leads to Type errors since we're trying to use modified Node structures with frozen metadata.

PR #642, opened 2026-07-26
perf: narrow precompile workload

MilesCranmerBot:
## Summary

- precompile the `Float32`, single-output search used by PySR defaults
- stop eagerly compiling the `Float64` and multi-output variants

## Why

The current workload runs a Cartesian grid of four searches during package precompilation. PySR uses 32-bit precision and a single target by default.

The scalar `equation_search` entry point immediately reshapes its target to `1 x N`. Both scalar and multi-output calls then enter the same `Vector{Dataset}` search implementation. Since a Julia `Vector` does not specialize on its length, the two-output workload repeats most of the core compilation.

Local clean-depot measurements on Julia 1.10.11:

- root package precompile time decreased from about 43 seconds to about 17 seconds
- root compiled cache decreased from about 59 MiB to about 39 MiB
- default PySR first-fit timings showed no repeatable difference beyond machine noise

Float64 and multi-output users will compile their remaining specializations on first use.

Related to #363.

## Checks

- `TEST_GROUP=unit/precompile ... Pkg.test()`: 10/10 passed
- `do_precompilation(Val(:compile))`: passed
- JuliaFormatter 2.4.0: clean
- repeated PySR default first-fit smoke benchmarks using the intended SymbolicRegression source path


MilesCranmer:
@MilesCranmerBot can you make this something that gets controlled by Preferences.jl please.

For example, the default should be to have (Float32, Float64) and (1,).

But we can have a Preferences flag `precompile_float64` which, if set to false, will disable the Float64 precompilation.

MilesCranmer (approved):

PR #608, opened 2026-05-24
Re-tune all hyperparameters for v2

MilesCranmer:
## Summary

This PR updates the default hyperparameters used by `Options()` and `defaults >= v"2.0.0-"`. The previous profile remains available through `defaults=v"1.0.0"`.

`niterations` is unchanged. Each iteration performs more evolutionary work, so the default search runs longer.

## Tuning benchmark

The hyperparameters were selected using a fixed suite of exact equations, noisy equations, feature-selection problems, and real regression datasets.

| Task | Target or dataset | Train/test | Operators | `maxsize` | Score |
|---|---|---:|---|---:|---|
| Damped oscillator | `2.3 exp(-t / 3.0) cos(1.7t + 0.4)` with Gaussian noise `σ = 0.05 std(y)` | 4096/4096 | `+ - * / exp cos` | 30 | Normalized Pareto volume |
| Smooth Friedman equation | `sin(π x₀ x₁) + (x₂ - 0.5)²` | 4096/4096 | `+ - * / sin` | 30 | Soft exact recovery, true complexity 13 |
| Feature-selection stress test | `sin(x₀) + x₁x₂`, with `x₃,…,x₉` as irrelevant inputs | 4096/4096 | `+ - * / sin` | 50 | Soft exact recovery, true complexity 7 |
| Log/square-root interaction | `log(x₀) / sqrt(x₁ + x₂)` | 4096/4096 | `+ - * / log sqrt` | 30 | Soft exact recovery, true complexity 9 |
| Noisy Friedman-1 | `10 sin(π x₀ x₁) + 20(x₂ - 0.5)² + 10x₃ + 5x₄`, with `x₅,…,x₉` irrelevant and Gaussian noise `σ = 0.1 std(y)` | 4096/4096 | `+ - * / sin` | 50 | Normalized Pareto volume |
| Dense linear regression | `Σᵢ wᵢxᵢ`, with 30 relevant features, fixed `wᵢ ~ N(0,1)`, and Gaussian noise `σ = 0.05 std(y)` | 4096/4096 | `+ - * /` | 100 | Normalized Pareto volume |
| California housing | Standardized 10,000-example subset of California housing | 8000/2000 | `+ - * / sin cos exp log sqrt tanh` | 50 | Normalized log-MSE improvement |
| ESL | Standardized PMLB `1027_ESL` | 390/98 | `+ - * / sin cos exp log sqrt tanh` | 30 | Normalized log-MSE improvement |
[...]

PR #643, opened 2026-07-26
Increase crossover probability in v2 defaults

MilesCranmerBot:
## Summary

This replaces the broad default changes proposed in #608 with a conservative v2 profile. Current and explicit v2 defaults inherit the v1 hyperparameters and change only `crossover_probability`, from `0.0259` to `0.20`. This keeps `adaptive_parsimony_scaling=1040.0`.

The benchmark constructors are pinned to the v1 profile so future default changes do not alter the benchmark workload.

## Evidence

The crossover-only profile was evaluated in a 560-search factorial ablation across 10 tuning tasks and 8 held-out regression datasets using 300-second searches and repeated seeds. It improved aggregate held-out Pareto-curve NMSE by 2.24% relative to the v1 profile.

A separate 420-search crossover sweep compared `0.0259`, `0.10`, `0.15`, `0.20`, `0.30`, and `0.40`. The `0.20` setting was the only alternative that improved aggregate held-out Pareto-curve NMSE, by 0.82%, and improved 6 of the 8 held-out datasets.

Lowering `alpha` or changing `adaptive_parsimony_scaling` to `20.0` worsened held-out performance and removed the crossover-only gain, so those values remain unchanged from v1.


MilesCranmer:
@MilesCranmerBot could we please leave the adaptive parsimony scaling at 20.0? I think >1000 is too harsh, even if it does do better at toy benchmarks.

MilesCranmer:
@MilesCranmerBot there are branch conflicts, please fix.

MilesCranmer:
@MilesCranmerBot formatting fix please.

src/Options.jl
```diff
@@ -1235,7 +1235,7 @@ function default_options(@nospecialize(version::Union{VersionNumber,Nothing} = n
         # Working with Complexities
         parsimony=0.0,
         warmup_maxsize_by=0.0,
-        adaptive_parsimony_scaling=1040.0,
+        adaptive_parsimony_scaling=20.0,
```
MilesCranmer: ????

@MilesCranmerBot you cannot change the v1 defaults... Please undo this. adaptive_parsimony_scaling is only 20.0 on v2!

MilesCranmerBot: Fixed in d557e59a: v1 base default restored to `adaptive_parsimony_scaling=1040.0`; the 20.0 override now applies only in the v2 branch alongside `crossover_probability=0.20`. Test updated to assert v1=1040.0 / v2=20.0. Verified locally against `default_options` for v1, v2, and unversioned.

MilesCranmer:
@MilesCranmerBot formatting... again...

MilesCranmer (approved):

MilesCranmer:
@MilesCranmerBot ugh I keep going back and forth on adaptive_parsimony_scaling. Can you undo the change and leave 1040 please? Back to where before

MilesCranmer:
@MilesCranmerBot this remaining CI error seems to be due to a caching weirdness. See https://github.com/SymbolicML/DynamicExpressions.jl/pull/182/commits/70b79dfb4218e20587e7cfd910538fb2c0a31fc9 for how you fixed it in the other thread. Try that fix here too. But ideally you should do this in a separate PR, ok? A separate PR for SR.jl that disables the cache ONLY FOR LoopVectorization tests. b/c LV tests are CPU sensitive.

MilesCranmer (approved):

src/Options.jl
MilesCranmerBot: You were right — the merged code still left the v2 value at `1040.0` despite the earlier acknowledgement. I opened #682 to restore `adaptive_parsimony_scaling=20.0` for v2/current defaults while preserving v1 at `1040.0`.

PR #645, opened 2026-07-27
feat!: adaptive-mutation / annealing / mutation-loop plugins + per-call mutation contexts

MilesCranmerBot:
Rewrites #612 on top of latest master, which absorbed #609 and #610. Supersedes #612.


MilesCranmer:
@codex please review

src/plugins/SimulatedAnnealing.jl
```diff
@@ -0,0 +1,105 @@
...
+end
+
+function on_cycle_start!(
+    s::SimulatedAnnealingState,
+    ::SimulatedAnnealingPlugin,
+    cycle_idx::Int,
+    options::AbstractOptions,
+)
+    n = options.ncycles_per_iteration
+    s.temperature = n > 1 ? (n - cycle_idx) / (n - 1) : 1.0
```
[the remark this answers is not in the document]
MilesCranmerBot: Fixed in 2dc7eb9b. on_cycle_start! now receives the actual ncycles passed to s_r_cycle, and simulated annealing uses that value. Added coverage for explicit three-cycle and one-cycle runs.

src/RegularizedEvolution.jl
```diff
@@ -25,16 +25,29 @@ function reg_evol_cycle(
...
+                    plugin_states,
+                    population_for_backsolve=pop,
+                )
+            wrapped_step = base_step
+            for i in min(length(options.plugins), length(plugin_states)):-1:1
+                inner = wrapped_step
+                plugin, pstate = options.plugins[i], plugin_states[i]
+                wrapped_step = parent -> wrap_mutation_step(pstate, plugin, parent, inner)
+            end
+            baby, mutation_accepted, tmp_num_evals = wrapped_step(allstar)
```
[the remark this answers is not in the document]
MilesCranmerBot: Fixed in 2dc7eb9b. Recorder-enabled mutation steps now keep a recorder per call and write the accepted compound lineage one parent/child edge at a time, while excluding rejected retry branches. Added a three-step compound recording test.

src/plugins/AdaptiveMutationWeights.jl
```diff
@@ -0,0 +1,151 @@
...
+    s::AdaptiveMutationWeightsState,
+    p::AdaptiveMutationWeightsPlugin,
+    mutation::AbstractMutation,
+    event::MutationEvent,
+    dataset,
+    options::AbstractOptions,
+)
+    skip_in_adaptive_weights(mutation) && return nothing
+    mutations = options.mutations
+    idx = findfirst(pair -> typeof(pair.first) === typeof(mutation), mutations)
```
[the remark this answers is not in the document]
MilesCranmerBot: Fixed in 2dc7eb9b. Attribution now matches the sampled configured mutation instance instead of its type. Added a test with two ConstantMutation instances using different perturbation factors.

src/plugins/AdaptiveMutationWeights.jl
```diff
@@ -0,0 +1,151 @@
...
+
+Whether [`AdaptiveMutationWeightsPlugin`](@ref) should exclude a mutation
+kind from its success/attempt accounting. Default `false`. Overridden to
+`true` for `SimplifyMutation` and `DoNothingMutation` (they accept trivially / don't
+represent real search moves).
+
+Extend by dispatch:
+
+```julia
+SymbolicRegression.skip_in_adaptive_weights(::MyMutation) = true
```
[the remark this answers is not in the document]
MilesCranmerBot: Fixed in 2dc7eb9b. skip_in_adaptive_weights is now public at SymbolicRegression.skip_in_adaptive_weights, with a third-party dispatch-extension test.

src/MutationFunctions.jl
```diff
@@ -119,20 +125,73 @@ function mutate_operator(
...
+    ctx::ConstantMutationContext,
+    m::ConstantMutation,
+    options::AbstractOptions,
+    rng::AbstractRNG=default_rng(),
+) where {T<:DATA_TYPE}
+    if !(has_constants(tree))
+        return tree
+    end
+    node = rand(rng, NodeSampler(; tree, filter=t -> (t.degree == 0 && t.constant)))
+    node.val = mutate_value(rng, node.val, ctx, m)
```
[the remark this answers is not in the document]
MilesCranmerBot: Fixed in 2dc7eb9b. Custom values continue through the public mutate_value(rng, value, temperature, options) contract; the context and mutation overload remains internal to numeric constants. Restored the custom String example and added a regression test.

src/Mutate.jl
```diff
@@ -320,11 +321,45 @@ end
...
+    curmaxsize::Int,
+    options::AbstractOptions;
+    tmp_recorder::RecordType,
+    plugin_states::Tuple=(),
+    population_for_backsolve=nothing,
+)::Tuple{
+    P,Bool,Float64
+} where {T,L,D<:Dataset{T,L},N<:AbstractExpression{T},P<:AbstractPopMember{T,L,N}}
+    for (plugin, pstate) in zip(options.plugins, plugin_states)
+        set_temperature!(pstate, plugin, temperature)
```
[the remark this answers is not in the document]
MilesCranmerBot: Fixed in 2dc7eb9b. The positional overload now initializes missing plugin state, forwards the supplied temperature to every opt-in plugin, and preserves legacy constant scaling when no plugin consumes it. Added annealing, non-annealing, and multiple-consumer regression tests.

benchmark/benchmarks.jl
```diff
@@ -94,15 +94,27 @@ function create_utils_benchmark()
     # Switch on `hasfield(Options, :plugins)` so this script benchmarks
     # cleanly against both versions.
     new_engine_sig = hasfield(Options, :plugins)
+    has_temperature = any(m -> m.nargs == 6, methods(next_generation))
```
MilesCranmer: Looks really fragile. Probably better to gate this based on package version?

benchmark/benchmarks.jl
```diff
@@ -171,6 +190,62 @@ function create_utils_benchmark()
                 ]
             )
         )
+    elseif new_engine_sig && !has_temperature
```
MilesCranmer: Can this be more DRY? Why do we need to write it all out again?

src/Core.jl
```diff
@@ -104,8 +105,17 @@ using .PluginModule:
...
     tournament_cost_multiplier,
     mutation_acceptance_multiplier,
+    MutationAcceptanceContext,
     fork_plugin_state,
+    refresh_plugin_state,
     resolve_init_member,
-    default_adaptive_parsimony_plugin
+    MutationStepResult,
+    wraps_mutation_step,
+    wrap_mutation_step,
```
MilesCranmer: Why have both? Seems like a code smell.

src/Mutate.jl
```diff
@@ -269,19 +273,20 @@ let mutation_types = BUILTIN_MUTATION_TYPES
...
     tmp_recorder::RecordType,
-    plugin_states::Tuple=(),
+    plugin_states::Tuple,
     population_for_backsolve=nothing,
 )::Tuple{
     P,Bool,Float64
 } where {T,L,D<:Dataset{T,L},N<:AbstractExpression{T},P<:AbstractPopMember{T,L,N}}
+    length(options.plugins) == length(plugin_states) || throw(
+        ArgumentError("`options.plugins` and `plugin_states` must have the same length."),
+    )
```
MilesCranmer: Why is this an issue? Surely that would never not be true?

src/Mutate.jl
```diff
@@ -293,20 +298,21 @@ end
     weights = copy(options.mutations)
 
     condition_mutation_weights!(weights, member, options, curmaxsize, nfeatures)
-    for (plugin, pstate) in zip(options.plugins, plugin_states)
+    map(options.plugins, plugin_states) do plugin, pstate
```
MilesCranmer: This looks nice. Can you do this in all contexts please? i.e., use this pattern instead of the `zip` everywhere we loop over plugins and states in the code?

src/Mutate.jl
```diff
@@ -350,6 +356,13 @@ function _next_generation(
     max_attempts = 10
     node_storage = allocate_container(member.tree)
 
+    mut_context = prepare_mutation_context(mutation_choice)
+    if mut_context !== nothing
```
MilesCranmer: Prefer `!isnothing` to `!== nothing`

src/MutationFunctions.jl
```diff
@@ -153,17 +144,28 @@ function mutate_constant(
...
+    rng::AbstractRNG,
+    val::Number,
+    temperature,
+    m::ConstantMutation,
+    options::AbstractOptions,
+)
+    return mutate_value(rng, val, temperature, m)
+end
+
+function _mutate_value(  # COV_EXCL_LINE
```
MilesCranmer: Why COV_EXCL_LINE needed here? Seems surprising...

src/Plugin.jl
```diff
@@ -170,12 +171,12 @@ function on_generation_end!(
...
 
 """
-    on_cycle_end!(plugin, state, pop, dataset, hof, options)
+    on_cycle_end!(state, plugin, pop, dataset, hof, options)
 
-Lifecycle hook called on the WORKER after each s_r_cycle finishes. May run
-concurrently across workers. Use only worker-local state, or use `Channel` /
-`RemoteChannel` for cross-worker communication. Called once per plugin per
-worker cycle.
+Lifecycle hook called on the WORKER at the end of each evolution cycle, paired
```
MilesCranmer: Why is `WORKER` capitalized?

src/Plugin.jl
```diff
@@ -280,6 +310,20 @@ function fork_plugin_state(head_state, ::AbstractPlugin, dataset)
...
+Refresh a persistent per-population worker state from the current head state
+before its next dispatch. The default preserves `worker_state`. Plugins whose
+head-side hooks update data consumed by workers can return a merged or replaced
+state.
+
+!!! warning "Experimental"
+"""
+function refresh_plugin_state(worker_state, head_state, ::AbstractPlugin, dataset)
+    return worker_state
+end
```
MilesCranmer, line 338: A bit unclear naming scheme. Also what gets refreshed vs what is used to refresh.

src/Population.jl
```diff
@@ -55,15 +55,17 @@ end
 function _init_tree(
     dataset, options, nlength::Int, nfeatures::Int, ::Type{T}, plugin_states::Tuple
 ) where {T}
-    fallback = gen_random_tree(nlength, options, nfeatures, T)
     candidate = resolve_init_member(plugin_states, options.plugins, dataset, options)
-    return isnothing(candidate) ? fallback : candidate::typeof(fallback)
+    if isnothing(candidate)
+        return gen_random_tree(nlength, options, nfeatures, T)
+    end
+    return candidate
```
MilesCranmer: Prefer `@something`

src/Population.jl
```diff
@@ -74,8 +76,11 @@ function Population(
     nlength::Int=3,
     nfeatures::Int,
     npop=nothing,
-    plugin_states::Tuple=(),
+    plugin_states::Tuple,
 ) where {T,L}
+    length(options.plugins) == length(plugin_states) || throw(
+        ArgumentError("`options.plugins` and `plugin_states` must have the same length."),
+    )
```
MilesCranmer: Delete?

src/Population.jl
```diff
@@ -133,7 +139,11 @@ Create random population and score them on the dataset.
     dataset = Dataset(X, y, L)
     update_baseline_loss!(dataset, options)
     return Population(
-        dataset; population_size=population_size, options=options, nfeatures=nfeatures
+        dataset;
+        population_size=population_size,
+        options=options,
+        nfeatures=nfeatures,
+        plugin_states,
```
MilesCranmer: don't need `=` here.

src/Population.jl
```diff
@@ -154,13 +164,16 @@ end
 
 # Sample the population, and get the best member from that sample
 function best_of_sample(
-    pop::Population{T,L,N}, options::AbstractOptions; plugin_states::Tuple=()
+    pop::Population{T,L,N}, options::AbstractOptions; plugin_states::Tuple
 ) where {T,L,N}
+    length(options.plugins) == length(plugin_states) || throw(
+        ArgumentError("`options.plugins` and `plugin_states` must have the same length."),
+    )
```
MilesCranmer: See other comments on patterns like this.

src/RegularizedEvolution.jl
```diff
@@ -1,54 +1,171 @@
...
+    return MutationStepLayer(state, plugin, inner)
+end
+@noinline function _plugin_state_mismatch()
+    throw(ArgumentError("`options.plugins` and `plugin_states` must have the same length."))
+end
+
+mutable struct MutationStep{D,P,O,S,H,A,M,R}
+    dataset::D
+    population::P
+    curmaxsize::Int
```
MilesCranmer: Could be better documented? I am a little confused. Also not sure why they are `mutable`. In Julia you can have an immutable that contains mutables, remember.

src/SingleIteration.jl
```diff
@@ -23,35 +29,35 @@ function s_r_cycle(
...
     best_examples_seen = HallOfFame(options, dataset)
     num_evals = 0.0
 
     batched_dataset = options.batching ? batch(dataset, options.batch_size) : dataset
 
-    for temperature in all_temperatures
+    for cycle_idx in 1:ncycles
+        for (plugin, pstate) in zip(options.plugins, plugin_states)
+            on_cycle_start!(pstate, plugin, cycle_idx, ncycles, options)
+        end
```
MilesCranmer: See comment earlier about switching to `map`

MilesCranmer (changes requested):
Some initial comments

test/unit/expressions/test_expression_builder.jl
```diff
@@ -130,10 +130,10 @@ end
...
 
-    @test_deprecated mutate_constant(copy(ex), 1.0, options, MersenneTwister(1))
+    mutate_constant(copy(ex), 1.0, options, ConstantMutation(), MersenneTwister(1))
 
     tree = Node(Float64; val=1.0)
     expression = Expression(tree; operators=options.operators)
-    @test_deprecated mutate_constant(copy(tree), 1.0, options, MersenneTwister(1))
-    @test_deprecated mutate_constant(copy(expression), 1.0, options, MersenneTwister(1))
+    mutate_constant(copy(tree), 1.0, options, ConstantMutation(), MersenneTwister(1))
+    mutate_constant(copy(expression), 1.0, options, ConstantMutation(), MersenneTwister(1))
```
MilesCranmer: I don't think these are necessary, are they? I thought they were just there to exercise the deprecated paths... which no longer exist

test/unit/evolution-core/test_prob_pick_first.jl
```diff
@@ -17,6 +17,11 @@
 
     for reverse in [false, true]
         T = Float32
+        dataset = Dataset(zeros(T, 1, n), zeros(T, n))
+        plugin_states = map(
+            plugin -> SymbolicRegression.init_plugin_state(plugin, options, dataset),
+            options.plugins,
+        )
```
MilesCranmer: We seem to do this a lot... How about making a function like `init_plugin_states(options, dataset)` that does this? (And implementing it through the library too?

src/plugins/MutationBurst.jl
```diff
@@ -0,0 +1,72 @@
...
+        n_steps = 1
+        while n_steps < p.compound_max_steps && rand() < p.compound_probability
+            next_result = next_step(result.member)
+            next_result.accepted || break
+            result = next_result
+            n_steps += 1
+        end
+        return result
+    end
+end
```
MilesCranmer: I think it would be cleaner to return a callable object rather than a closure. There shouldn't be any reason this is  `@unstable`.

MilesCranmer (changes requested):

MilesCranmer (approved):

MilesCranmer:
@codex review

src/SymbolicRegression.jl
```diff
@@ -711,8 +730,36 @@ end
...
+        WorkerPluginStatesType[
+            if j == 1 && i == 1
+                worker_plugin_state_prototype
+            else
+                strictmap(
+                    (plugin, head_state) ->
+                        fork_plugin_state(head_state, plugin, datasets[j]),
+                    options.plugins,
+                    plugin_states[j],
+                )::WorkerPluginStatesType
```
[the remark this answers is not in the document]
MilesCranmer: This is overkill; can ignore

MilesCranmer (approved):

PR #646, opened 2026-07-28
fix: finalize logging after all search outputs complete

adil-soubki:
## Summary

- Continue processing remaining outputs when one output reaches its final cycle.
- Invoke `logging_callback!` after every completed cycle, including the terminal cycle.
- Emit the default logger's terminal event exactly once after all outputs finish.
- Add regression coverage for terminal callbacks and multi-output searches.

## Root cause

The search loop used `break` when the current output reached zero remaining cycles. In a multi-output search, that exited the shared loop as soon as one output completed and could leave other outputs unfinished. The callback also ran only on the nonterminal path, so loggers could miss the final state.

The loop now skips rescheduling completed outputs while continuing to process the others. Logging runs after each completed cycle, and the default logger identifies the global terminal state from all outputs.

## Validation

- `unit/basic`: 1494 assertions
- `unit/parallel`: 13 assertions
- Multi-output multithreading probe observed one intermediate state and one terminal state

### DIFF
--- a/src/Logging.jl
+++ b/src/Logging.jl
@@ -21,9 +21,10 @@ Abstract type for symbolic regression loggers. Subtypes must implement:
 
 - `get_logger(logger)`: Return the underlying logger
 - `logging_callback!(logger; kws...)`: Callback function for logging.
-    Called with the current state, datasets, runtime options, and options. If you wish to
-    reduce the logging frequency, you can increment and monitor a counter within this
-    function.
+    Called once after each completed search cycle, including the globally terminal
+    cycle, with the current state, datasets, runtime options, and options. If you wish
+    to reduce the logging frequency, you can increment and monitor a counter within
+    this function.
 """
 abstract type AbstractSRLogger <: AbstractLogger end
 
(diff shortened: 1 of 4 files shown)


MilesCranmer:
@codex review

adil-soubki:
Updated in b210bf72 and rebased onto current master.

Good catch on the multi-output case: the existing `break` exited the entire search when the first output finished. The loop now reschedules only outputs with remaining cycles, allowing every output to complete. The callback runs once per completed cycle, including exactly once for the globally terminal state.

`SRLogger` now guarantees a terminal emission for positive `log_interval` values without duplicating it when the interval already aligns. Nonpositive intervals remain disabled.

The large-looking loop diff is mostly indentation around unchanged scheduling code. New regressions cover multi-output completion, aligned and unaligned intervals, no duplicate terminal emission, disabled logging, and TensorBoard counts.

Local verification passed: 1,494 basic tests, 13 parallel tests, 58 MLJ/TensorBoard integration tests, and JuliaFormatter 2.4.0.

adil-soubki:
The failing JET test in CI seems unrelated?

MilesCranmer:
@codex review

MilesCranmer (approved):

MilesCranmer:
@MilesCranmerBot please backport to release-v1

PR #648, opened 2026-07-29
Rework backsolve mutation into a monotone budget-aware sparse fit

MilesCranmerBot:
## Summary

Builds on the experimental `BacksolveMutation` so the SINDy-style move delivers its intended value: invert a random subtree's context per data row, fit a replacement subtree by greedy forward selection over a population-wide basis library under the remaining complexity budget, and only accept children whose cost on the full dataset beats the parent's. The mutation remains off by default (`backsolve::Float64 = 0.0`).

@MilesCranmer says:
> @ayagh19 @adil-soubki
>
> I was messing around with backsolve trying to figure out why enabling it didn't seem to give as much of a search improvement as I'd hoped. Here is what a few days worth of analysing results and messing around with coding agents found! It seems to work MUCH better now.

## Background

Backsolve is meant to be a SINDy-like move inside the GA: compute the values a subtree should have for the whole expression to match the target, then solve for a better subtree directly. In the initial implementation, enabling it regressed benchmarks by 2-3 orders of magnitude. The mechanics behind that, all measured:

- The inversion failed entirely if any single data row left the inverse's domain (e.g. log of a negative value). ~86% of accepted children were unchanged clones of the parent, and since accepted children displace the oldest population member, the mutation was actively homogenizing the population.
- Fitted replacements usually violated `curmaxsize`, so the retry loop rebuilt and refit the full library up to 10 times per event before giving up: ~500ms per event vs ~4ms for a normal mutation, with none of those evaluations charged to `num_evals`.
- Replacements discarded the existing subtree (or fell back to a bare constant when the fit failed), and children were accepted unconditionally: only ~7.5% of accepted children improved on their parent.

## Changes

- **Row-masked inverse evaluation** (`eval_inverse_tree_array_masked`): rows where the surrounding context is not invertible are excluded from the fit instead of failing the whole inversion. Their targets are pinned to the current subtree's values, preserving current behavior where the inversion gives no information.
- **Budget-aware greedy forward selection** (replacing STLSQ): terms are added by residual correlation with per-term complexity costs, so results respect the remaining `curmaxsize` budget by construction. Fits share one precomputed Gram matrix across attempts; least-squares refits use a Cholesky solve with a small Tikhonov ridge fallback.
- **Monotone acceptance**: a child is returned only if its cost on the full dataset (re-evaluated on the current batch when batching) improves on the parent's. Otherwise the event reports failure via a new `success` field on `MutationResult` and is skipped.
- **Library from the whole population**, deduplicated by evaluated value (previously the top 10 members only), plus the subtree being replaced, so fits can retain existing structure.
- Guards for unsupported expression wrappers, non-binary node types, missing targets, and operator sets without `+` and `*` now report mutation failure instead of throwing.
- Internal evaluations are charged to `num_evals`.

`BacksolveMutation` configuration changes (experimental API): `lambda` and `max_iter` are replaced by `max_terms`, `min_improvement`, and `node_attempts`.

## Measured results

Event-level, before vs after:

| | before | this PR |
|---|---|---|
| accepted children identical to parent | 86% | 0% |
| accepted children worse than parent | ~50% | 0% (monotone gate) |
| accepted events that improve parent cost | ~7.5% | 71-91% |
| wall time per backsolve event | ~533ms | ~2-8ms |

End-to-end (backsolve at weight 1.0, 45s wall, 3 seeds per task):

- Noisy linear regression with 30/100/300 features: previously regressed 400-3600x vs baseline; now matches baseline at the noise floor on every seed.
- Nonlinear and additive-assembly exact-recovery tasks: previously failed all seeds (loss ~1.1-4.7); now recovers exactly (loss < 1e-30) on every seed.
- 30-term polynomial with `maxsize=150`: baseline solves 2/3 seeds with median 146s; this PR solves 3/3 with median 27s (~5x faster), recovering exact monomial sums with least-squares coefficients.


MilesCranmer:
@codex review

MilesCranmer:
@codex review

MilesCranmer (approved):

PR #668, opened 2026-08-08
compat: adopt DynamicExpressions 2.9 EvalContext

MilesCranmerBot:
## Summary

- require DynamicExpressions 2.9
- adopt `EvalContext` and `eval_context` throughout evaluation and buffered template paths
- retain `EvalOptions` as a deprecated SymbolicRegression binding

## Why

DynamicExpressions 2.9 renamed the public evaluation keyword from `eval_options` to `eval_context`, matching the `EvalContext` type and its call-scoped mutable buffer state.

## Validation

- registered DynamicExpressions 2.9.0 smoke test passed
- `unit/templates`: 380 passed
- `unit/evolution-core`: 1,188 passed
- `unit/expressions`: 145 passed
- JuliaFormatter 2.1.6 and `git diff --check`


MilesCranmer:
@MilesCranmerBot please update to DE v2.9. Now we can pass `eval_context` (with `eval_options` as a deprecated option also still available). Let's update to this please.

src/InterfaceDynamicExpressions.jl
```diff
@@ -64,25 +64,25 @@ which speed up evaluation significantly.
...
                 )
             else
-                eval_options
+                eval_context
             end
-            (; eval_options=de_eval_options,)
+            (; eval_context=de_eval_context,)
         else
             NamedTuple()
         end
```
MilesCranmer, line 105: I'm not sure this is backwards compatible? What happens if I provide `eval_options` explicitly here. Does it get overwritten? Or are there two conflicting args?

MilesCranmer: And this concern repeats in other areas too. `eval_options` should be a depwarn but still work. Just like DE itself handles.

MilesCranmer (changes requested):
@MilesCranmerBot please address

MilesCranmer:
@codex review

src/ComposableExpression.jl
```diff
@@ -64,16 +65,18 @@ end
     tree::AbstractExpressionNode{T};
     operators::Union{AbstractOperatorEnum,Nothing}=nothing,
     variable_names::Union{AbstractVector{<:AbstractString},Nothing}=nothing,
-    eval_options::Union{Nothing,EvalOptions}=nothing,
+    eval_context::Union{Nothing,EvalContext}=nothing,
+    eval_options::Union{Nothing,EvalContext}=nothing,
 ) where {T}
-    if eval_options !== nothing && eval_options.buffer !== nothing
+    eval_context = _process_eval_options(eval_context, eval_options, :ComposableExpression)
```
MilesCranmer: It would be better if we put deprecated kwargs into a generic `kws...` and then checked it. You can see other examples of deprecated keywords throughout the library.

MilesCranmer (changes requested):
@MilesCranmerBot please address

MilesCranmer:
@codex review

src/LossFunctions.jl
```diff
@@ -165,8 +170,10 @@ function eval_loss(
     options::AbstractOptions;
     regularization::Bool=true,
     idx=nothing,
-    eval_options=nothing,
+    eval_context=nothing,
+    kws...,
```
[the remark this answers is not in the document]
MilesCranmer: @MilesCranmerBot please fix this in a new PR

PR #700, opened 2026-08-18
refactor: replace StatsBase dependency with in-house StatsBaseLite

MilesCranmerBot:
## What

StatsBase was only used for four things:

- weighted sampling (`sample(items, Weights)` / `sample(Weights)`) in `MutationWeights.jl` and `Population.jl`
- sampling without replacement (`sample(rng, 1:n, k; replace=false)`) in `TemplateExpression.jl` and `Population.jl`
- `mean` in `SearchUtils.jl` (this is just `Statistics.mean` reexported)

This PR adds `src/third_party/StatsBaseLite.jl`, a vendored subset of StatsBase.jl v0.34 containing exactly the code paths behind those callsites, and drops StatsBase from `[deps]`.

The code is a **verbatim port** of the upstream algorithms (not a reimplementation): the `Weights` type, the `replace=false` poly-algorithm (`k == 1` single draw / `k == 2` `samplepair` / `n < 24k` Fisher-Yates / `self_avoid_sample!`), `direct_sample!`, and the cumulative-scan weighted single draw. The StatsBase MIT license and copyright notice are retained in the file header. `ordered=true` (sequential sampling), multi-draw weighted sampling, and the AliasTables path are deliberately omitted and throw or are absent, since no callsite uses them.

The module is included inside the `@stable` block, so DispatchDoctor instruments it under the same test-time stability checking as the rest of the package.

## Why

Dependency reduction for load/precompile time: StatsBase itself costs ~4.6s of fresh precompile and pulls in SortingAlgorithms, Missings, StatsAPI, DataAPI, AliasTables. This PR removes it from SymbolicRegression's direct dependency graph, so SR's own code and precompiled images no longer reference StatsBase types.

**Caveat:** Optim 1.13.3 still hard-depends on StatsBase, so until the planned follow-up vendoring of Optim lands, StatsBase (~9ms warm load) and SparseArrays (~169ms warm load) still enter the runtime load graph transitively through Optim. The user-visible load/precompile saving materializes with that follow-up; this PR is the prerequisite step.

## Compatibility

No public API change: StatsBase was imported qualified (`using StatsBase: StatsBase`), never reexported, and no user-facing signature accepts StatsBase types. Internal callsites now alias the vendored module (`using ..StatsBaseLite: StatsBaseLite as StatsBase`), so the callsites are unchanged.

## Verification

- `Pkg.test()` with TEST_GROUP = `unit/basic`, `unit/plugins`, `unit/templates`, `unit/evolution-core`: all pass (exit 0)
- New `test/unit/basic/test_statsbase_lite.jl` verifies **seeded exact parity** with real StatsBase 0.34 (test-only dependency): identical draws across every poly-algorithm branch, weighted draws including zero-weight exclusion, plus contract tests (no-replacement uniqueness, `k > n` error, non-finite weight rejection). 25/25 pass.
- `Pkg.resolve()`: StatsBase pruned from SR's direct dependencies in a re-resolved Manifest
- `Base.@time_imports using SymbolicRegression`: SR itself no longer imports StatsBase; StatsBase/SparseArrays still appear via Optim as noted in the caveat above


MilesCranmer:
Turns out Optim.jl also pulls in StatsBase, so we don't even win anything with this. Sigh.

PR #704, opened 2026-08-21
feat: stop searches gracefully on interrupt via stop flag or pipe fd

MilesCranmerBot:
## Summary

Adds a per-search `ExternalStop` object to the existing early-stop check in the master search loop.

- `ExternalStop.requested` is an atomic Boolean for direct stop requests.
- On POSIX systems, `ExternalStop` can poll a nonblocking pipe descriptor. Only the configured trigger byte requests a stop.
- The stop state travels through `RuntimeOptions`, so concurrent and sequential searches do not share cancellation state.
- The search drains queued notifications, latches a request for the current search, and tears down normally at a cycle boundary.

A disabled stop remains the default and specializes to a no-op.

## Motivation

PySR embeds Julia through PythonCall. A SIGINT delivered directly into the embedded Julia runtime can raise an asynchronous `InterruptException` inside an arbitrary scheduled task. If that task holds a runtime lock, the exception can corrupt lock accounting or terminate the host process.

This API lets a host request cooperative cancellation through search-local state. The search returns equations found so far through the normal completion path.

PySR integration: astroautomata/PySR#1310.

## Concurrency and lifecycle

Each `equation_search` call owns its `ExternalStop`. Overlapping searches therefore stop independently. Before entering the search loop, the implementation drains and latches any request already queued for that object. During the search, pipe polling is throttled to a 20 ms minimum interval to avoid paying for a system call on every scheduler check.

## Testing

- Added unit coverage for direct requests, matching and unrelated trigger bytes, queued requests before search entry, descriptor draining, default no-op behavior, and overlapping serial searches.
- The focused `test_external_stop.jl` suite passes: 28 tests.
- GitHub Actions passes on `format-check`, `Documentation`, `CI`, and `CI Extended`. The extended matrix completed all 30 jobs successfully.
- End-to-end PySR tests pass against this branch, including real subprocess SIGINT, real Jupyter kernel interruption, partial results, kernel survival, and a second fit in the same process.


MilesCranmer:
@codex review

MilesCranmer:
@codex review

MilesCranmer:
@codex review

src/SymbolicRegression.jl
```diff
@@ -1300,6 +1305,7 @@ function _main_search_loop!(
             check_for_user_quit(state.stdin_reader),
             check_for_timeout(start_time, options),
             check_max_evals(state.num_evals, options),
+            check_external_stop(),
```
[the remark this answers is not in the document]
MilesCranmerBot: Breaking here drops the same in-flight outputs as the existing early-stop conditions in this tuple: `check_for_loss_threshold`, `check_for_user_quit`, `check_for_timeout`, and `check_max_evals` all break at this exact statement with identical semantics. Ready-but-unharvested channels hold at most one cycle per worker, and the hall of fame already contains those populations' results from their previous harvest, so the marginal loss is one evolution cycle. Making stop requests drain ready channels while the sibling conditions do not would give one condition special semantics; if draining is wanted it should cover all five conditions as its own change.

MilesCranmer:
@codex review

MilesCranmer:
@codex review

src/SearchUtils.jl
```diff
@@ -462,6 +468,100 @@ function check_max_evals(num_evals, options::AbstractOptions)::Bool
...
+            end
+        end
+    end
+    return stop.requested[]
+end
+
+@inline latch_external_stop!(::Nothing)::Bool = false
+@inline latch_external_stop!(ropt::AbstractRuntimeOptions)::Bool = latch_external_stop!(
+    ropt.external_stop
+)
```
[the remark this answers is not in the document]
MilesCranmer: Fixed in 901f2c07: added an `external_stop(::AbstractRuntimeOptions) = nothing` accessor (overridden for `RuntimeOptions`), so custom subtypes default to no external stop and can opt in by overriding. Covered by a new testitem with a fieldless subtype.

MilesCranmer:
@codex review

src/SymbolicRegression.jl
```diff
@@ -690,6 +697,8 @@ end
     saved_state,
     guesses,
 ) where {D<:Dataset}
+    # Preserve a request queued while the `X, y` overload prepared its datasets.
```
MilesCranmer: Delete

src/SymbolicRegression.jl
```diff
@@ -1184,7 +1194,7 @@ function _main_search_loop!(
             ###################################################################
 
             state.cycles_remaining[j] -= 1
-            if state.cycles_remaining[j] > 0
+            if state.cycles_remaining[j] > 0 && !check_external_stop(ropt)
```
MilesCranmer, line 1196: Just wondering - why do we give special treatment to this type of stop, but not for, e.g., timeout or user quit?

src/SymbolicRegression.jl
```diff
@@ -1337,6 +1348,9 @@ function _tear_down!(
         # TODO: We should unwrap the error monitors here
         state.we_created_procs && rmprocs(state.procs)
     end
+    # Consume notifications that arrived after the loop's final poll (e.g.
+    # while waiting on workers) so they cannot stop a later search.
```
MilesCranmer: Noisy; delete

test/unit/misc/test_external_stop.jl
MilesCranmer, line 1: Just to confirm - does this actually get executed? Do we not need to include this file explicitly somewhere? (I forget)

src/SymbolicRegression.jl
MilesCranmer: Corrected answer (previous version overstated the latency): the real distinction is latched vs consuming checks, not urgency.

- `check_external_stop` *latches* (`requested[]` stays true), so it is safe to consult in both the refill guard and the bottom break: consuming it here cannot lose the stop.
- `check_for_user_quit` *consumes* stdin bytes and returns an edge. If it were added to this guard, a quit detected here would be swallowed: the bottom break would re-poll an empty stream and the loop would keep running with that population never rescheduled.
- Timeout, loss threshold, and max-evals are level-triggered and could safely be shared, but omitting them here only costs at most one extra scheduled cycle for the just-arrived population before the bottom break catches it (this loop handles one ready result per iteration).

So the guard includes exactly the check that is both latched and represents an interrupt where skipping that one extra cycle matters. Uniform treatment would want a single `should_stop` computed once per iteration (latching the quit edge into state), which is a clean follow-up refactor but out of scope here.

src/SearchUtils.jl
```diff
@@ -462,6 +468,108 @@ function check_max_evals(num_evals, options::AbstractOptions)::Bool
...
+
+"""Check for an external stop request and latch matching pipe notifications."""
+function check_external_stop(stop::ExternalStop)::Bool
+    if stop.fd >= 0
+        now = time_ns()
+        if now - stop.last_poll_ns[] >= STOP_POLL_INTERVAL_NS
+            stop.last_poll_ns[] = now
+            # Keep draining even once latched so notifications arriving after
+            # the first stop cannot remain queued on the owned descriptor.
+            if check_stop_fd(stop)
```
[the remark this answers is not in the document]
MilesCranmerBot: Fixed in ea1a9eb4: `_tear_down!` now ends with a `drain_external_stop!` call that consumes notifications queued after the loop's final poll (e.g. while waiting on in-flight workers), so they cannot stop a later search sharing the descriptor. Requests arriving after teardown are still preserved for the next search. Covered by a testitem that drives `_tear_down!` directly with a queued byte (c657b0ec).

src/SymbolicRegression.jl
MilesCranmerBot: Removed in a0758027.

src/SymbolicRegression.jl
MilesCranmerBot: Removed in a0758027.

src/SymbolicRegression.jl
MilesCranmerBot: Corrected answer (previous version overstated the latency): the real distinction is latched vs consuming checks, not urgency.

- `check_external_stop` *latches* (`requested[]` stays true), so it is safe to consult in both the refill guard and the bottom break: consuming it here cannot lose the stop.
- `check_for_user_quit` *consumes* stdin bytes and returns an edge. If it were added to this guard, a quit detected here would be swallowed: the bottom break would re-poll an empty stream and the loop would keep running with that population never rescheduled.
- Timeout, loss threshold, and max-evals are level-triggered and could safely be shared, but omitting them here only costs at most one extra scheduled cycle for the just-arrived population before the bottom break catches it (this loop handles one ready result per iteration).

So the guard includes exactly the check that is both latched and represents an interrupt where skipping that one extra cycle matters. Uniform treatment would want a single `should_stop` computed once per iteration (latching the quit edge into state), which is a clean follow-up refactor but out of scope here.

test/unit/misc/test_external_stop.jl
MilesCranmerBot: Yes: `test/runtests.jl` uses TestItemRunner's `@run_package_tests`, which discovers `@testitem` blocks by scanning the package (no explicit include needed) and filters on `ti.filename` starting with `test/<TEST_GROUP>`. CI.yml's unit matrix includes `misc`, so `unit/misc` runs this file on all OS/Julia combinations; the local runs above also picked up all 5 testitems this way.

MilesCranmer (approved):

PR #724, opened 2026-09-16
docs: fix Cambridge versioned deployment

MilesCranmerBot:
## Problem

Every tagged release build for the Cambridge docs target (ai.damtp.cam.ac.uk/symbolicregression) was deployed into `/dev/` instead of `/vX.Y.Z/`. `Documenter.deploy_folder` was evaluated while `GITHUB_REPOSITORY` still named `astroautomata/SymbolicRegression.jl`, so the repository check failed, the deploy decision had an empty subfolder, `siteinfo.js` was written with `DOCUMENTER_CURRENT_VERSION = ""`, and the later `deploydocs` call (which did have the override) pushed the release into `dev`. The version picker therefore showed only `dev`, and `stable` still pointed at v1.12.0.

## Changes

- Apply the Cambridge `GITHUB_REPOSITORY` / `DOCUMENTER_KEY` override before `auto_detect_deploy_system()` and `deploy_folder`, with one consistent repo string.
- `keep=:patch` so `vX.Y.Z` folders are deployed alongside `stable`, `vX`, `vX.Y`. PySR pins its backend to patch precision and links to the matching folder.
- Filter "Migrating from v1" out of the top nav in `config.mts`; the page stays in the sidebar.

## Verification

Simulated tag build (`GITHUB_REF=refs/tags/v2.4.1`, `DEPLOYMENT_TARGET=cambridge`): every deployment criterion passes and bases are `v2.4.1, stable, v2, v2.4`. The v2.4.1 build from this branch has been deployed to the Cambridge site; `/symbolicregression/v2.4.1/` is live with `DOCUMENTER_CURRENT_VERSION = "v2.4.1"` and the nav no longer contains the migration entry.


MilesCranmer (approved):

MilesCranmer:
Reopened as https://github.com/astroautomata/SymbolicRegression.jl/pull/725; this PR was tied to a branch that was removed.

PR #726, opened 2026-09-20
perf: run population copy and hall-of-fame output as plain loops

MilesCranmerBot:
`Base.copy(::Population)` and `save_to_file` ran their loops under `Threads.@threads` on the head thread. Each call spawns one task per thread, and `copy` runs twice per population per iteration, so at 96 threads and 96 populations the head thread launches thousands of tasks per iteration that compete with the worker tasks. A head-thread profile at 96 threads spent 1990 of 2015 samples inside `threading_run` waking threads.

Per-iteration wall on a 96-core Genoa node, 96 populations x 27 members, plain `Node`, Julia 1.12, `parallelism=:multithreading`:

| threads | before | after |
|---|---|---|
| 1 | 26 ms | 26 ms |
| 48 | 104 ms | 88 ms |
| 96 | 2271 ms | 189 ms |

The default search (96 x 27, 100 iterations) goes from 196 s to 17 s at 96 threads. Pareto front and best loss are unchanged. A laptop sees little change from this.

Tests: full `unit/` suite locally on Julia 1.12 with this and the follow-up allocation changes applied.


MilesCranmer (commented):
If benchmark is good, seems like a good fix

MilesCranmer (approved):

PR #727, opened 2026-09-20
perf: run optimize_and_simplify_population as a plain loop

MilesCranmerBot:
`optimize_and_simplify_population` ran its per-member loop under `Threads.@threads` unless `deterministic` or Enzyme was set. With `parallelism=:multithreading` it is already called from inside a worker task, one per population, so every thread is busy and the inner `@threads` only spawns `nthreads` more tasks per population per iteration. The loop is now plain, `@threads_if` had no other user and is removed, and `AutoEnzyme` is no longer needed in `SingleIteration.jl`.

Serial searches lose the inner threading over `population_size` members, which was the only place a serial search used threads.

Default preset (96 populations x 27, 100 iterations, 96 threads, Genoa node, Julia 1.12): wall 15.9 s to 14.8 s, peak RSS 3.9 GB to 1.9 GB. Pareto front and best loss are unchanged.


MilesCranmer:
@codex review

MilesCranmer (approved):

PR #729, opened 2026-09-20
perf: splice crossover subtrees without copying them

MilesCranmerBot:
`crossover_trees` copies both parent trees, then copies each chosen subtree again before splicing it into the other parent. The second copy is not needed for `Node`: each subtree is spliced exactly once, `n2` leaves `t2` when `n1` takes its place, and `n1` has already left `t1` by then. Only a root swap keeps the whole tree, and that path still copies. For node types with `preserve_sharing` (`GraphNode`), a subtree can have other parents, so both are still copied.

This removes two subtree copies per crossover. Measured on the default preset (96 populations x 27, 100 iterations, 96 threads, Genoa node) with an arena-backed node type where subtree copies are the dominant cost, wall dropped 15.0 s to 13.9 s; for `Node` the saving is the two copies' worth of allocation per crossover. Pareto front and best loss are unchanged.


MilesCranmer:
@codex review

PR #730, opened 2026-09-20
perf: hoist s_r_cycle plugin hook closures into functions

MilesCranmerBot:
The plugin `on_cycle_start!`/`on_cycle_end!` calls in `s_r_cycle` were `do`-block closures inside the cycle loop. `pop` is reassigned in that loop, so Julia boxes it, and each closure re-captures `options` on every cycle. Moving the two calls into named functions removes the box and the per-cycle closure allocation (about 1.3 KB per cycle).

Default preset (96 populations x 27, 100 iterations, 96 threads, Genoa node, Julia 1.12): bytes allocated 48.6 GB to 45.1 GB; wall within noise. Pareto front and best loss are unchanged.


MilesCranmer:
@codex review

MilesCranmer (approved):

PR #731, opened 2026-09-20
perf: reuse a per-task scratch for conditioned mutation weights

MilesCranmerBot:
`next_generation` copies `options.mutations` on every mutation so `condition_mutation_weights!` can adjust it in place. The copy is a fresh `Vector{Pair{AbstractMutation,Float64}}` per mutation.

Now the conditioned weights are written into a per-task scratch vector (`PerTaskCache`, the same mechanism `Utils.jl` already uses for `SafeFunctions`), resized and `copyto!`-filled from `options.mutations` each time. Behavior is unchanged; the scratch is never read across tasks.

Measured together with the tournament-selection change (separate PR) on the default preset (96 populations x 27, 100 iterations, 96 threads, Genoa node, Julia 1.12): wall 14.8 s to 11.1 s, bytes allocated 43 GB to 29 GB; this change is the smaller share of that.


MilesCranmer:
@codex review

MilesCranmer (approved):

PR #732, opened 2026-09-20
perf: reuse one evaluation buffer per population slot across iterations

MilesCranmerBot:
`s_r_cycle` calls `create_eval_context`, which allocates `curmaxsize` row-length arrays, on every population on every iteration. Those buffers are the evaluator's scratch and are dead once the cycle returns.

`SearchState` now owns one `EvalContext` per population slot (`eval_contexts[j][i]`), created empty, passed down through `_dispatch_s_r_cycle` to `s_r_cycle`, and grown in place with `grow_eval_context!` the first time a cycle needs more arrays than it holds. Only one task per slot runs at a time, so nothing is shared between tasks. With `parallelism=:multiprocessing` the slot holds `nothing` and the worker creates its own context as before, since the buffer cannot cross processes. Options with `bumper=true` or operators that take no context keep `nothing` everywhere.

Default preset (96 populations x 27, 100 iterations, `Node`, 96 threads on a 128-core Rusty node, Julia 1.12), 3 reps each, with the sibling perf PRs (#727 to #731) applied to both arms:

| | wall | bytes allocated | peak RSS |
|---|---|---|---|
| before | 17.6 ± 0.6 s | 24.9 GB | 1.98 GB |
| after | 15.6 ± 0.2 s | 22.3 GB | 1.78 GB |

Pareto front and best loss are identical.

Tests: full `unit/` suite locally on Julia 1.12 (34761 tests), which covers the batching, turbo, bumper, and multiprocessing groups.


MilesCranmer:
@codex review

